PR #7, opened 2016-06-05
Clamp log knob values

zamaudio:
Signed-off-by: Damien Zammit <damien@zamaudio.com>

### DIFF
--- a/dgl/src/ImageWidgets.cpp
+++ b/dgl/src/ImageWidgets.cpp
@@ -647,14 +647,14 @@ float ImageKnob::_logscale(float value) const
 {
     const float b = std::log(fMaximum/fMinimum)/(fMaximum-fMinimum);
     const float a = fMaximum/std::exp(fMaximum*b);
-    return a * std::exp(b*value);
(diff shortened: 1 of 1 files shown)


zamaudio:
My last PR introduced this bug, so here's a fix.  Alternatively you can revert my last PR since I created a new widget with the needed changes instead.

falkTX:
I agree with the change, but not how it was changed.
Use std::min(maxcalc, realmax) instead of comparing and setting maxcalc value.

zamaudio:
Updated PR

falkTX:
hmm wait, so the issue here is that 'value' variable is out of bounds?
seems to me that the bug is somewhere else.

falkTX:
I think a better solution here is just comparing 'value' against minimum and maximum.
like:

```
if (value <= fMinimum)
    return value;
if (value >= fMaximum)
    return value;
// rest of the code, now safe to calculate log
```

zamaudio:
This bug was introduced because of my last PR, can you revert it please? I don't even need it.

falkTX:
hmm ok :)

PR #10, opened 2016-06-13
Separate 64 bit builds from 32 bit builds of WIN

zamaudio:
Signed-off-by: Damien Zammit <damien@zamaudio.com>


falkTX:
This is completely unnecessary.
Just add "-m32" to your CFLAGS, CXXFLAGS and LDFLAGS, or call the 32bit compiler.

PR #17, opened 2016-10-30
Feature/windows xpile 2

guysherman:
- Added ability to display an error message when trying to open lv2 dlls on Windows (in lv2_ttl_generator).
- Added a windows batch file analogous to generate-ttl.sh

### DIFF
--- a/utils/generate-ttl.bat
+++ b/utils/generate-ttl.bat
@@ -1,0 +1,18 @@
+@echo off
+IF NOT EXIST bin GOTO NOBINDIR
+CD bin
+SET GEN=%~dp0\lv2_ttl_generator.exe
+SET EXT=dll
+for /D %%s in (.\*) do (call :generatettls "%%s")
+
+
+goto :eof
+:NOBINDIR
+echo "Please run this script from the surface root"
+goto :eof
+
+:generatettls
+pushd %1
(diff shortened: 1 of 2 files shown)


falkTX:
Can you tell me why this is needed?
Are you building the plugins with MSVC?

(also, your lv2_ttl_generator.c changes are missing an #ifdef)

guysherman:
I'll fix the .c file - bad copy paste :(

I have been working on cross compiling from the Windows Subsytem for Linux (essentially Ubuntu 14.04, but running on Windows). 

The one problem was that you can't run the lvw_ttl_generator.exe from the bash console (because unlike msys2, it is actually a linux environment), so you have to run it from the windows command-prompt after you finish compiling (I can submit a PR to your sample repository as well, to fold in the changes from the 'consumer' side). I wrote a batch file to do the same job as the shell script. 

The changes to lv2_ttl_generator are helpful for tracking down dll hell.

falkTX:
Adding that code for windows will require linking to windows libraries.
I'll have to make it optional.

guysherman:
I didn't have to add any link flags.

falkTX:
afaik the ubuntu-for-windows thing can now run windows binaries, so this PR might not be needed anymore.

falkTX:
I am closing this issue due to not being needed anymore, and not something I am supporting right now.
Thanks anyway.

Issue #19, opened 2017-07-25
Add support for effEditKeyDown and effEditKeyUp


fundamental:
Some VST hosts do not reliably relay keyboard events to the plugin, with one notable example being FL Studio. From researching this issue it appears that some (deprecated) VST opcodes are used to send the messages instead. I would ask that effEditKeyDown (59) and effEditKeyUp (60) are added to support keyboard events under these hosts. 

Information is scattered around the net about these particular opcodes (one example for reference https://forum.juce.com/t/vst-plugin-still-not-getting-keystrokes/6582/14 ). So, it should be possible to add this without too much of a headache.

falkTX:
I did a quick commit on DPF in order to support this.
https://github.com/DISTRHO/DPF/commit/fa54750583225c2e6aa5cda9baf9c18eb722daee

Completely untested. I don't have enough time to dig deeper into this, but that should give you enough to work with for now.

fundamental:
Thanks. I'll add this to the windows build and I'll report back if the commit is successful or not.

fundamental:
It looks like most of the patch works, though there are some remaining issues.

- Duplicate key presses are observed in hosts which support both effEditKeyDown and normal window events. I'd suggest ignoring effEditKeyDown events when events are observed via the normal window interface. (Reaper is an example of an affected host)
- Special Key presses (notably shift and control) are not handled correctly. Shift for example is incorrectly interpreted as a space key press.
- The effEditKeyDown/Up commit breaks the build for VST plugins which use an external/out-of-process UI (i.e. zyn's ntk UI). I would expect the correct way to handle this would to just be to disable the extra event handling in the out-of-process UI case.

See https://www.kvraudio.com/forum/viewtopic.php?t=488363&p=6839472#p6839472 for the original user feedback.

Thanks again for the fast implementation.

falkTX:
There should be a way to detect if the host uses those vst keys or not, perhaps a can-do return...
If not, we just hardcode a whitelist of hosts and keep adding more as needed.
Will have to investigate later.

Considering I did no tests on this code at all (only that it builds), I'm surprised some things actually work :)
There's a few things I believe you're able to fix yourself maybe,
if you can then please do and send a PR to DPF later.

SpotlightKid:
This can be closed, can it not? PR #20 is closed.

falkTX:
it does not work fully right yet. plus we have to start whitelisting good known hosts.
right now keys are getting duplicated in some of them.

[...]

PR #20, opened 2017-07-30
Vst key events

fundamental:
- Fixes Ctrl/Shift/Alt event handling
- Fixes build for Zyn's ntk external UI

See also: https://github.com/DISTRHO/DPF/issues/19

For zyn-fusion I've internally implemented a fix for the repeated key presses. With these changes I'd say the vst-key-event handling should be good enough to merge into master, though that's your call in the end.

If nothing else merging these with the DISTRHO/DPF repo should make it easy for me to get out some more builds with the fixes with the build-from-scratch scripts I have.

### DIFF
--- a/distrho/src/DistrhoPluginVST.cpp
+++ b/distrho/src/DistrhoPluginVST.cpp
@@ -187,6 +187,9 @@ public:
         case  6: index = kCharEscape;    break;
         case 22: index = kCharDelete;    break;
 
+        //Other keycodes
+        case 7:  index = ' '; break;
+
         // handle rest of special keys
         case 40: special = kKeyF1;       break;
         case 41: special = kKeyF2;       break;
@@ -214,7 +217,7 @@ public:
         if (special != 0)
             return fUI.handlePluginSpecial(down, static_cast<Key>(special));
 
-        if (index > 0 || (value == 7 && index == 0))
+        if (index > 0)
             return fUI.handlePluginKeyboard(down, static_cast<uint>(index));
 # endif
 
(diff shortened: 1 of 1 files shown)


fundamental:
@falkTX would you mind reposting your question? I don't see it in this PR or in any of the commits associated with this PR.

falkTX:
In line 216, `if (index > 0 || (value == 7 && index == 0))`
> index is 0 here, which means no key pressed right?
> what is the reason for this change?

fundamental:
When pressing space in reaper I see index=0, value=7. All other keys that I have looked at the debug output seem to indicate that index=0 generally means special (non-numeric/alphabetic/basic symbol key).

falkTX:
I understand that, but right now the code will be calling handlePluginKeyboard() with a key id 0 (index variable is the one passed to that function).
I suggest addline a line right after https://github.com/fundamental/DPF/blob/32239a031399cbb900f69e9c561635887416734d/distrho/src/DistrhoPluginVST.cpp#L186
like this:
```
case 7: index = ' '; break;
```
what do you think?

fundamental:
Sounds fine to me and it appears to work in my testing setup. The fix has been added to this PR.

falkTX:
I imported the commit that was clean, and did the rest myself (wanted to reword some comments there).
I also added a print for the host name when the UI is started, and put the code in place to start the white-list.
This is only needed on windows right? or is macOS affected too?
We should run the plugin in the affected hosts, listen to the printed messages and add to the whitelist.
The host-name print will remain to make it easier to add new names to the list later.

PR #28, opened 2018-01-07
Add kParameterIsTrigger which maps to pprops:trigger

grejppi:
Just a little cosmetic thing I was missing when porting some existing plugins to DPF 😊

### DIFF
--- a/distrho/DistrhoPlugin.hpp
+++ b/distrho/DistrhoPlugin.hpp
@@ -89,6 +89,12 @@ static const uint32_t kParameterIsLogarithmic = 0x08;
  */
 static const uint32_t kParameterIsOutput = 0x10;
 
+/**
+   Parameter value is a trigger.
+   This is only supported in LV2 plugins.
+*/
+static const uint32_t kParameterIsTrigger = 0x20;
+
 /** @} */
 
 /* ------------------------------------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


falkTX:
I was going to ask about if it was LV2 specific, but you mentioned it in the comment for the property.
For standalones, we can do it without issues, there's no host involved.
It's possible to do it for VSTs too, though in a kinda hacky way (the entire VST spec is hacky anyway).

LADSPA and DSSI cannot do this.

So I like the idea, but I'd change things a little:
1. set value as `0x20 | kParameterIsBoolean`, to make sure it always includes boolean prop.
2. change comment to mention it breaks LADSPA and DSSI
3. implement in JACK and VST formats

If you want to help more, go ahead and do the first 2 items.
Thanks.

grejppi:
> 1. set value as 0x20 | kParameterIsBoolean, to make sure it always includes boolean prop.

Done ☺️ 

> 2. change comment to mention it breaks LADSPA and DSSI

I'm still a little confused about this part, but I agree, the LV2 specification seems to require that the value is reset immediately.

The way I do it in my plugin is along the lines of:
```cpp
float parameter;
bool triggered;

void setParameterValue(uint32_t index, float value) override {
    // ...
    parameter = value;
        if (value > 0.5 && !triggered) {
            triggered = true;
            // do something once
        } else {
            triggered = false;
        }
    }
    // ...
}
```

A user of the LADSPA version would need to set the control to 0 manually before it can be triggered again, but because the value is only updated when it goes from 0 to 1 doesn't make any difference however long the value remains 1.

falkTX:
Now implemented in DPF. Added a fake mode of it for ladspa, vst and jack formats.

PR #29, opened 2018-01-07
Explicitly state supported LV2 options & use param:sampleRate

grejppi:
Semantically [`lv2:sampleRate`](http://lv2plug.in/ns/lv2core/#sampleRate) is a port property (equivalent to `LADSPA_HINT_SAMPLE_RATE`), not a parameter. Jalv, for example, uses [`param:sampleRate`](http://lv2plug.in/ns/ext/parameters/#sampleRate) with a float value for this purpose. ([See here](https://github.com/drobilla/jalv/blob/5f015f74427e39ca411032f5f42bc969f89cf9fe/src/jalv.c#L1077-L1078))

### DIFF
--- a/distrho/src/DistrhoPluginLV2export.cpp
+++ b/distrho/src/DistrhoPluginLV2export.cpp
@@ -259,7 +259,6 @@ void lv2_generate_ttl(const char* const basename)
         pluginString += "    opts:supportedOption <" LV2_BUF_SIZE__nominalBlockLength "> ,\n";
         pluginString += "                         <" LV2_BUF_SIZE__maxBlockLength "> ,\n";
         pluginString += "                         <" LV2_PARAMETERS__sampleRate "> ,\n";
-        pluginString += "                         <" LV2_CORE__sampleRate "> ;\n";
         pluginString += "\n";
 
         // UI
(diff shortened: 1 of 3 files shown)


distrho/src/DistrhoPluginLV2export.cpp
```diff
@@ -254,6 +255,13 @@ void lv2_generate_ttl(const char* const basename)
 #endif
         pluginString += ";\n\n";
 
+        // supportedOptions
+        pluginString += "    opts:supportedOption <" LV2_BUF_SIZE__nominalBlockLength "> ";
```
falkTX: these lines don't follow the rest of the syntax on this file.
leave the `,\n` at the end.

falkTX:
Part of this was handled in https://github.com/DISTRHO/DPF/commit/68b3ca4abe46ee7386cc20574964d1c62c0c1e21
I forgot the DSP side, only fixed UI parts.

Using LV2_CORE__sampleRate as property is wrong.
I will remove that from LV2 DSP side.

PR looks good, except for 1 small syntax thing and that "core" property which we should remove.

falkTX:
Nice, thanks.
If you remove the LV2_CORE__sampleRate usage, I'll merge right away :)

distrho/src/DistrhoPluginLV2export.cpp
```diff
@@ -259,7 +259,6 @@ void lv2_generate_ttl(const char* const basename)
         pluginString += "    opts:supportedOption <" LV2_BUF_SIZE__nominalBlockLength "> ,\n";
         pluginString += "                         <" LV2_BUF_SIZE__maxBlockLength "> ,\n";
         pluginString += "                         <" LV2_PARAMETERS__sampleRate "> ,\n";
```
falkTX: TTL line needs to end with `;`

PR #32, opened 2018-01-28
Fix Circle<T>::getY()

desaulniersp:


falkTX:
nice catch! thank you

PR #34, opened 2018-02-11
Concat DGL_NAMESPACE properly in pugl osx

pdesaulniers:


falkTX:
I noticed the macros were broken some days ago too..
but I'd rather not include that file like what you suggest here.

I will take this as a bug report :)

PR #37, opened 2018-02-24
Proper namespace for isBase64Char

pdesaulniers:


falkTX:
hah, it's obvious where this code came from :)

PR #38, opened 2018-02-25
OSX VST: Remove dependence on SetFile

### LINKED ISSUE #36: Remove dependence on `SetFile` in OSX VST builds
Idea taken from here: 

https://github.com/overdrivenpotato/rust-vst2/blob/master/osx_vst_bundler.sh

:+1:

Isn't this the only remaining step to cross compile without needing a physical OSX machine?

zamaudio:
Fixes #36

### DIFF
--- a/utils/generate-vst-bundles.sh
+++ b/utils/generate-vst-bundles.sh
@@ -11,11 +11,6 @@ fi
 
 PWD=`pwd`
 
-if [ ! -d /System/Library ]; then
-  echo "This doesn't seem to be OSX, please stop!"
-  exit 0
-fi
-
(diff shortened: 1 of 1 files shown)


zamaudio:
Tested on OSX 10.6.8 (bundles appear with VST icon).

zamaudio:
The force pushed commit above is a better way, since there is no reason to call SetFile (tested on 10.6.8) anymore for some reason...

falkTX:
Sorry it took so long. I don't like to boot into MacOS if I can avoid it.
Confirmed to work fine without SetFile.

Issue #27, opened 2017-11-28
Set VST parameter hints where useful


falkTX:

pdesaulniers:
Can be closed now. :)

falkTX:
yes, indeed :)

PR #39, opened 2018-02-25
Set VST parameter hints where useful

pdesaulniers:
https://github.com/DISTRHO/DPF/issues/27

Hold on :p This has some issues in Carla, which, if I understand correctly, expects a fixed and unnormalized value when the parameter uses IntStep. It worked as expected in Bitwig, though, and Renoise + Ardour would just ignore the flag. Perhaps there's something I don't understand?

I tweaked the code so that the parameter would work as expected in Carla, but then it wouldn't get past 1.0f in other hosts. I might have made a dumb mistake :)

Anyway, I'm really tired, so I'm off to bed. I'll take a closer look at this during the day.

### DIFF
--- a/distrho/src/DistrhoPluginVST.cpp
+++ b/distrho/src/DistrhoPluginVST.cpp
@@ -468,7 +468,7 @@ public:
                 }
                 else if (hints & kParameterIsInteger)
                 {
-                    DISTRHO_NAMESPACE::snprintf_iparam((char*)ptr, (int32_t)(value + 0.5f), 24);
+                    DISTRHO_NAMESPACE::snprintf_iparam((char*)ptr, (int32_t)std::round(value), 24);
                 }
                 else
                 {
--- a/distrho/src/DistrhoPluginVST.cpp
+++ b/distrho/src/DistrhoPluginVST.cpp
@@ -451,17 +451,17 @@ public:
         case effGetParamDisplay:
             if (ptr != nullptr && index < static_cast<int32_t>(fPlugin.getParameterCount()))
             {
-                uint32_t hints = fPlugin.getParameterHints(index);
+                const uint32_t hints = fPlugin.getParameterHints(index);
                 const float value = fPlugin.getParameterValue(index);
 
-                if(hints & kParameterIsBoolean)
+                if (hints & kParameterIsBoolean)
                 {
                     const ParameterRanges& ranges(fPlugin.getParameterRanges(index));
                     const float midRange = ranges.min + (ranges.max - ranges.min) / 2.0f;
                     
-                    DISTRHO_NAMESPACE::strncpy((char*)ptr, value > midRange ? "on" : "off", 4);
+                    DISTRHO_NAMESPACE::snprintf_param((char*)ptr, value > midRange ? ranges.max : ranges.min, 24);
                 }
-                else if(hints & kParameterIsInteger)
+                else if (hints & kParameterIsInteger)
                 {
                     DISTRHO_NAMESPACE::snprintf_param((char*)ptr, (int32_t)value, 24);
                 }
@@ -1083,16 +1083,16 @@ static intptr_t vst_dispatcherCallback(AEffect* effect, int32_t opcode, int32_t 
         }
         return 0;
     case effGetParameterProperties:
-        if(ptr != nullptr && index < static_cast<int32_t>(plugin.getParameterCount()))
+        if (ptr != nullptr && index < static_cast<int32_t>(plugin.getParameterCount()))
         {
             const uint32_t hints = plugin.getParameterHints(index);
             
-            if(hints & kParameterIsOutput)
+            if (hints & kParameterIsOutput)
                 return 1;
 
-            if(VstParameterProperties* const properties = (VstParameterProperties*)ptr) 
+            if (VstParameterProperties* const properties = (VstParameterProperties*)ptr) 
             {
-                if(hints & kParameterIsInteger)
+                if (hints & kParameterIsInteger)
                 {
                     properties->flags = kVstParameterUsesIntStep | kVstParameterUsesIntegerMinMax;
                     const ParameterRanges& ranges(plugin.getParameterRanges(index));
@@ -1102,12 +1102,12 @@ static intptr_t vst_dispatcherCallback(AEffect* effect, int32_t opcode, int32_t 
                     properties->minInteger = static_cast<int32_t>(ranges.min);
                     properties->maxInteger = static_cast<int32_t>(ranges.max);
                 }
-                else if(hints & kParameterIsBoolean)
+                else if (hints & kParameterIsBoolean)
                 {
                     properties->flags = kVstParameterIsSwitch;
                 }
 
-                if(hints & kParameterIsLogarithmic)
+                if (hints & kParameterIsLogarithmic)
                 {
                     properties->flags |= kVstParameterCanRamp;
                 }
--- a/distrho/src/DistrhoPluginVST.cpp
+++ b/distrho/src/DistrhoPluginVST.cpp
@@ -452,14 +452,19 @@ public:
             if (ptr != nullptr && index < static_cast<int32_t>(fPlugin.getParameterCount()))
             {
                 const uint32_t hints = fPlugin.getParameterHints(index);
-                const float value = fPlugin.getParameterValue(index);
+                float value = fPlugin.getParameterValue(index);
 
                 if (hints & kParameterIsBoolean)
                 {
                     const ParameterRanges& ranges(fPlugin.getParameterRanges(index));
                     const float midRange = ranges.min + (ranges.max - ranges.min) / 2.0f;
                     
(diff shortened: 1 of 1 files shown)


falkTX:
I did not have any plugins that used these hints to test in Carla, so Carla might be broken here.
You saw #27, it's in my plans to try to get this working.

pdesaulniers:
The LSP plugins do, apparently. They seem to work fine, I'll take a look at its source code as soon as sourceforge stops hanging.

pdesaulniers:
So, the LSP plugins have the same problem if I compile them from source. The int parameters get stuck between 0 and 1, except in Carla, where they work as expected. 

Seems like the official builds just floor the float parameter and remove the decimal places in the display. The parameter doesn't step from value to value. It's boring, but it seems like that's the safest approach.

falkTX:
Looks good, but some notes:
1. use a space after an `if`
2. don't force a custom text render of boolean values, the "on" and "off".

some time later I will add scale-points to DPF, which we can use to render custom text to describe values in a better way. it's supported in LV2 and VST (via paramDisplay anyway).

falkTX:
Thanks for the changes.
Question regarding Carla, is this working properly there, or does Carla needs fixing to handle these flags properly?

pdesaulniers:
I think it needs some fixing to handle IntStep properly. Here is an example of a parameter with a range of 0 to 4 (sorry for the very bad screenshots):




The knob is also not completely aligned with the values:




If I hover the knob, it gets into its correct position:


Except for the misaligned knob, the flag works as expected if I send fixed values in the plugin's code (let's ignore parameters that aren't integers):
```c++
float vst_getParameter(const int32_t index)
{
    const ParameterRanges& ranges(fPlugin.getParameterRanges(index));
    //return ranges.getNormalizedValue(fPlugin.getParameterValue(index));
    return ranges.getFixedValue(fPlugin.getParameterValue(index));
}
```
```c++
void vst_setParameter(const int32_t index, const float value)
{
    const ParameterRanges& ranges(fPlugin.getParameterRanges(index));
    //const float realValue(ranges.getUnnormalizedValue(value));
    const float realValue(ranges.getFixedValue(value));

    fPlugin.setParameterValue(index, realValue);
    ...
}
```
```c++
void setParameterValue(const uint32_t index, const float realValue)
{
    const ParameterRanges& ranges(fPlugin->getParameterRanges(index));
    //const float perValue(ranges.getNormalizedValue(realValue));
    const float perValue(ranges.getFixedValue(realValue));

    fPlugin->setParameterValue(index, realValue);
    hostCallback(audioMasterAutomate, index, 0, nullptr, perValue);
}
```
But then, the int parameters get stuck between 0 and 1 in other hosts.

I hope that wasn't too much info :)

distrho/src/DistrhoPluginVST.cpp
```diff
@@ -1058,7 +1082,41 @@ static intptr_t vst_dispatcherCallback(AEffect* effect, int32_t opcode, int32_t
...
+                {
+                    properties->flags = kVstParameterUsesIntStep | kVstParameterUsesIntegerMinMax;
+                    const ParameterRanges& ranges(plugin.getParameterRanges(index));
+
+                    properties->stepInteger = 1;
+                    properties->largeStepInteger = 1;
+                    properties->minInteger = static_cast<int32_t>(ranges.min);
+                    properties->maxInteger = static_cast<int32_t>(ranges.max);
+                }
+                else if (hints & kParameterIsBoolean)
```
pdesaulniers: What about parameters that are both an int and a bool? Should I flip these ifs around? Or consider both flags, and print an int instead of a float in ParamDisplay?

distrho/src/DistrhoPluginVST.cpp
```diff
@@ -1058,7 +1082,41 @@ static intptr_t vst_dispatcherCallback(AEffect* effect, int32_t opcode, int32_t
...
+                {
+                    properties->flags = kVstParameterIsSwitch;
+                }
+
+                if (hints & kParameterIsLogarithmic)
+                {
+                    properties->flags |= kVstParameterCanRamp;
```
pdesaulniers: Is that "flags" field guaranteed to be zero initialized? I'm not very familiar with C/C++ yet so I'm not sure :p

falkTX: Good catch.
We should initialize the entire structure to zero after getting the pointer from the host.

distrho/src/DistrhoPluginVST.cpp
falkTX: ah yes, you should check for booleans first, then integers.

pdesaulniers:
It seems to me like there is some fixed value / normalized value confusion regarding int parameters in Carla. 

Here's a dumb workaround that makes it kind of work (CarlaPluginVST2.cpp at line 352):
```c++
    void setParameterValue(const uint32_t parameterId, const float value, const bool sendGui, const bool sendOsc, const bool sendCallback) noexcept override
    {
        CARLA_SAFE_ASSERT_RETURN(fEffect != nullptr,);
        CARLA_SAFE_ASSERT_RETURN(parameterId < pData->param.count,);

        float targetValue;

        if (pData->param.data[parameterId].hints & PARAMETER_IS_INTEGER) 
        {
            targetValue = pData->param.ranges[parameterId].getFixedAndNormalizedValue(value);
        }
        else
        {
            targetValue = pData->param.getFixedValue(parameterId, value);
        }

        fEffect->setParameter(fEffect, static_cast<int32_t>(parameterId), targetValue);

        CarlaPlugin::setParameterValue(parameterId, targetValue, sendGui, sendOsc, sendCallback);
    }
```

Same thing for setParameterValueRT. 

In audioMasterAutomate, instead of using getFixedValue all the time:
```c++
float targetValue;
            
if (pData->param.data[uindex].hints & PARAMETER_IS_INTEGER) 
{
    targetValue = pData->param.ranges[uindex].getUnnormalizedValue(opt);
}
else
{
    targetValue = pData->param.getFixedValue(uindex, opt);
}
```

There remains some issues afterwards, such as parameter reset not working properly, but it's the general idea.

Do you see what's going on? I'm not familiar enough with either Carla or VST, so I don't know how to fix this properly.
 
Also, it seems like there is a bug in the most recent Carla-git that isn't present in the beta6 release. The auto-generated sliders don't "snap" when moved using left-click, although they do when using the mouse wheel. This affects bool and int parameters.

falkTX:
I removed the integer min/max from carla now. I think it was wrong to begin with, very strange that vst has it for integers but not floats... I will just assume normalized values for everything.
Regarding the 2nd issue, please report a new bug against the carla project.

Here's a list things I ask for change:
1. the part with `hints & kParameterIsOutput` should be a few lines below, just after we `memset()` the host pointer.
2. the new `snprintf_param` should be renamed `snprintf_iparam` (added "i"), just my personal preference, avoiding overloaded functions.
3. converting floats to ints should be done like this, in order to round properly: `(int32_t)(value + 0.5f);`
4. the part with `properties->flags = kVstParameterIsSwitch` should be done independent of integer, like the ramp/log check is. a parameter can be integer and boolean at the same time, they're not mutually exclusive (we can also have switches which are not integers, of course)
4. remove kVstParameterUsesIntStep and the setting of step values. we're missing this info on DPF and the host can do it automatically anyway.

distrho/src/DistrhoPluginVST.cpp
```diff
@@ -463,7 +463,7 @@ class PluginVst
                 }
                 else if (hints & kParameterIsInteger)
                 {
-                    DISTRHO_NAMESPACE::snprintf_param((char*)ptr, (int32_t)value, 24);
+                    DISTRHO_NAMESPACE::snprintf_iparam((char*)ptr, (int32_t)(value + 0.5f), 24);
```
pdesaulniers: Actually, this won't work with negative parameters. Shouldn't this be (int32_t)std::round(value) ?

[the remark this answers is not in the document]
pdesaulniers: I ended up changing it for a round.

distrho/src/DistrhoPluginVST.cpp
```diff
@@ -452,14 +452,19 @@ class PluginVst
...
 
                 if (hints & kParameterIsBoolean)
                 {
                     const ParameterRanges& ranges(fPlugin.getParameterRanges(index));
                     const float midRange = ranges.min + (ranges.max - ranges.min) / 2.0f;
                     
-                    DISTRHO_NAMESPACE::snprintf_param((char*)ptr, value > midRange ? ranges.max : ranges.min, 24);
+                    value = value > midRange ? ranges.max : ranges.min;
+
+                    if (hints & kParameterIsInteger)
```
falkTX: this part here is not needed, and actually duplicated.
We can remove this if/else. then turn the `else if (hints & kParameterIsInteger)` check below into its own `if`, reducing the code a little.

falkTX:
All good now.
Hope you don't mind me nit-picking on the small things, take it as learning experience :)

Thanks a lot!

PR #44, opened 2018-04-20
Fix shaky TimeBeat in LV2 plugins

pdesaulniers:
Seems like the floating point => int32 conversion was causing some issues.

According to @rghvdberg, VST plugins also have that bug, but I've had some trouble reproducing it. 

In Ardour, I've found that the timebeat in the VST build of the "Info" plugin starts causing issues only if I open the "Plugin analysis" pane. It then twitches between 1 and its correct value. It starts working fine as soon as I close the pane. 

I've tested the VST build in other hosts, such as Carla, Renoise and Qtractor. The timebeat was correct.

The Jack build doesn't seem affected.

### DIFF
--- a/distrho/src/DistrhoPluginLV2.cpp
+++ b/distrho/src/DistrhoPluginLV2.cpp
@@ -387,7 +387,7 @@ public:
                     if (fLastPositionData.barBeat >= 0.0f)
                     {
                         const double rest = std::fmod(fLastPositionData.barBeat, 1.0f);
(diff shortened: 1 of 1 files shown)


falkTX:
Thanks.
I planned to take a look at this, but just fixing the transport in carla took longer than expected..
It is a lot harder than I first imagined.

I dont have time to test this now, I trust you that it works.
I will still do some of my own testing later on too.

PR #45, opened 2018-04-25
Add support for plugins that want midi input but aren't synths

pdesaulniers:
Plugins with DISTRHO_PLUGIN_WANT_MIDI_INPUT set to true and DISTRHO_PLUGIN_IS_SYNTH set to false would get the wrong run() method definition.

Is this fix correct?

### DIFF
--- a/distrho/src/DistrhoPluginCarla.cpp
+++ b/distrho/src/DistrhoPluginCarla.cpp
@@ -317,7 +317,7 @@ protected:
         fPlugin.deactivate();
     }
 
-#if DISTRHO_PLUGIN_IS_SYNTH
+#if DISTRHO_PLUGIN_WANT_MIDI_INPUT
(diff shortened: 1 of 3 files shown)


falkTX:
What about DSSI, LV2 and VST2 formats?

Also I noticed that macro has a typo, should be "WANTS_MIDI_INPUT"..
But I can handle this later one myself.

pdesaulniers:
Seems like they already use the correct macro.

https://github.com/DISTRHO/DPF/blob/master/distrho/src/DistrhoPluginVST.cpp#L861
https://github.com/DISTRHO/DPF/blob/master/distrho/src/DistrhoPluginLV2.cpp#L540
https://github.com/DISTRHO/DPF/blob/master/distrho/src/DistrhoPluginLADSPA%2BDSSI.cpp#L193

falkTX:
Thanks for the quick response and links.
I guess I started a transition to the new macro but did not complete it.
Will merge now, thank you.

PR #46, opened 2018-05-08
Handle paths with spaces when generating ttl

pdesaulniers:
The script would fail with a ```binary operator expected``` error at line 14.

Does this fix make sense? My bash skills are pretty abysmal :)

Also, the default Makefile for DPF plugins fails at gen if the path contains spaces due to an unquoted CURDIR: https://github.com/zamaudio/zam-plugins/blob/master/Makefile#L22

### DIFF
--- a/utils/generate-ttl.sh
+++ b/utils/generate-ttl.sh
@@ -9,13 +9,13 @@ else
   exit
 fi
 
-PWD=`dirname $0`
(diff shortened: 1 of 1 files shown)


falkTX:
Cool, I just wanted a confirmation. will merge now then.

PR #48, opened 2018-05-12
Support GNU platforms other than Linux

zamaudio:
Treat all glibc-based platforms as DISTRHO_OS_LINUX, even those not
based on Linux such as GNU/kFreeBSD or GNU/Hurd.

Also, wrap ifdefs around sa_restorer which really is specific to (some)
Linux platforms.  (A feature-test macro would be better, though it is
obsolete and maybe should be removed altogether...)


zamaudio:
Patch taken from https://salsa.debian.org/multimedia-team/zam-plugins/blob/master/debian/patches/0020-nonlinux.patch - compiles on GNU/hurd and GNU/kfreebsd

PR #49, opened 2018-05-19
Allow negative scaling of NanoVG context

pdesaulniers:
Negative scaling can be used for flipping an axis.


pdesaulniers:
It's a nice way of mirroring an image, or any drawing code. FWIW, memononen doesn't seem to be against it: https://github.com/memononen/nanovg/issues/351

It's also possible to pull this off using transform(), though.

PR #52, opened 2018-05-21
 Implement sendNote for VST2;  Add sendNote stub for Jack

pdesaulniers:
I've tested it, and it seems to work.

### DIFF
--- a/distrho/src/DistrhoPluginJack.cpp
+++ b/distrho/src/DistrhoPluginJack.cpp
@@ -93,7 +93,7 @@ public:
     PluginJack(jack_client_t* const client)
         : fPlugin(),
 #if DISTRHO_PLUGIN_HAS_UI
-          fUI(this, 0, nullptr, setParameterValueCallback, setStateCallback, nullptr, setSizeCallback, fPlugin.getInstancePointer()),
+          fUI(this, 0, nullptr, setParameterValueCallback, setStateCallback, sendNoteCallback, setSizeCallback, fPlugin.getInstancePointer()),
 #endif
           fClient(client)
     {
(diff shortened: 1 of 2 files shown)


distrho/src/DistrhoPluginJack.cpp
```diff
@@ -512,6 +523,13 @@ class PluginJack
     }
 #endif
 
+#if DISTRHO_PLUGIN_WANT_MIDI_INPUT
```
falkTX: sendNoteCallback is only defined if using a plugin with midi input, which means the constructor will fail on a plugin with different configuration :P

distrho/src/DistrhoPluginVST.cpp
```diff
@@ -280,8 +280,28 @@ class UIVst
...
+        midiEvent.midiData[1] = note;
+        midiEvent.midiData[2] = velocity;
+
+        VstEvents event;
+        memset(&event, 0, sizeof(VstEvents));
+
+        event.numEvents = 1;
+        event.events[0] = (VstEvent*)&midiEvent;
+
+        fEffect->dispatcher(fEffect, effProcessEvents, 0, 0, &event, 0.0f);
```
falkTX: This is nasty... effProcessEvents is supposed to be called by the plugin and only during processing...
Did you try this with carla? very likely carla will see it being called from the wrong thread.

It needs a different approach I believe.

distrho/src/DistrhoPluginJack.cpp
pdesaulniers: Oops. That wasn't very good :p

distrho/src/DistrhoPluginVST.cpp
pdesaulniers: Carla seems happy with it. 

Anyway, what would be a good approach?

falkTX: a ring buffer with a read-try-lock on the process/RT side of things.

dpf does not have such thing.
I dont think there will be a lot of events from the UI ever, as RT side can get them pretty quickly.
so perhaps just a static array of 16 events, with a mutex that does tryLock() on vst process would work fine.

pdesaulniers:
Anyway, this is a mess. Let's pretend the pull request never happened :D

falkTX:
You dont have to close it, you can work more on it, and then I squash the entire thing into a single commit. :)

PR #56, opened 2018-06-05
Return proper bounds in NanoVG::textBounds

pdesaulniers:
From the NanoVG header at https://github.com/DISTRHO/DPF/blob/master/dgl/src/nanovg/nanovg.h#L527 : 
```
The bounds value are [xmin,ymin, xmax,ymax]
```

Thus, b[2] - b[0] to get the width, b[3] - b[1] to get the height.


falkTX:
Makes sense. Thanks!

PR #64, opened 2018-08-22
Avoid gcc >= 8 warning about possible string truncation when using std::strncpy

### LINKED ISSUE #63: strncpy warnings with GCC 8
WIth GCC 8 I get the following warning on compilation:

```In file included from ../../dpf/distrho/DistrhoPluginMain.cpp:29:
In function 'void DISTRHO::strncpy(char*, const char*, size_t)',
    inlined from 'intptr_t DISTRHO::PluginVst::vst_dispatcher(int32_t, int32_t, intptr_t, void*, float)' at ../../dpf/distrho/src/DistrhoPluginVST.cpp:465:43:
../../dpf/distrho/src/DistrhoPluginVST.cpp:79:17: warning: 'char* strncpy(char*, const char*, size_t)' output may be truncated copying 23 bytes from a string of length 32 [-Wstringop-truncation]
     std::strncpy(dst, src, size-1);
     ~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~
```

Since the custom `DISTRHO::strncpy` function inserts a null char at `size-1` in any case, can't we just use `std::memcpy` instead of `std::strncpy` to get rid of the warning?

SpotlightKid:
Fixes #63.

### DIFF
--- a/distrho/src/DistrhoPluginVST.cpp
+++ b/distrho/src/DistrhoPluginVST.cpp
@@ -76,7 +76,7 @@ static const writeMidiFunc writeMidiCallback = nullptr;
 
 void strncpy(char* const dst, const char* const src, const size_t size)
 {
-    std::strncpy(dst, src, size-1);
+    std::memcpy(dst, src, std::min(std::strlen(src), size-1));
     dst[size-1] = '\0';
 }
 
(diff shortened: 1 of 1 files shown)


falkTX:
Not sure if you forgot.. but please apply the memcpy change to the other 2 functions, just as done above.
I will merge it right away.

You confirm this fixes the warning, right?

SpotlightKid:
I probably won't have tie to work on this until the week end.

falkTX:
I did not realize until now that the other 2 functions dont just copy bytes, but they are printf variants.
So what I said does not apply, the PR is good to go as-is.
Thank you.

PR #73, opened 2018-09-29
Add onPositionChanged callback for widgets

pdesaulniers:
This is useful for container widgets. Simple example:
```c++
void ParentWidget::onPositionChanged(const PositionChangedEvent &ev)
{
    fSubWidget->setAbsolutePos(ev.pos); //the subwidget moves with the parent
}
```

### DIFF
--- a/dgl/Widget.hpp
+++ b/dgl/Widget.hpp
@@ -176,6 +176,22 @@ public:
     };
 
    /**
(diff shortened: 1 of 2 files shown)


falkTX:
It is impossible to say no when the PR is so clean as this one...
I guess it won't hurt anyone.

Issue #68, opened 2018-09-13
Need clarification on NanoWidget constructor with NanoWidget* argument


pdesaulniers:
Namely, that one: https://github.com/DISTRHO/DPF/blob/4a4f95fc9f106882c2702fa22c2a574124a3f783/dgl/src/NanoVG.cpp#L945

It allows reusing an existing NanoVG context, which makes sense to me (I don't see the point in using multiple contexts?)

I've added an example to the gl-examples that shows its behavior: https://github.com/pdesaulniers/gl-examples/blob/master/examples/nanovg3.cpp

In the code, I set the absolute location of three rectangles to (100, 100), (200, 200) and (300, 300).

Right now, all the rectangles are getting drawn in the top left corner of the window. The position of the NanoRectangles is effectively ignored:
![2018-09-13-13 28 18](https://user-images.githubusercontent.com/23723605/45504829-e615e600-b758-11e8-895a-2c2c677b8186.png)

However, if I instantiate the rectangles with the `NanoWidget(Widget* groupWidget, int flags)` constructor instead, they appear at the expected location:
![2018-09-13-13 35 28](https://user-images.githubusercontent.com/23723605/45505238-23c73e80-b75a-11e8-9847-54ba43fcd559.png)

Is this the expected behavior?

falkTX:
This is clarified on the upcoming pugl-update-v2 branch by simply not having such constructor available.
Not the best, but at least not confusing.

We can later on do some tests to see if this could be made to work.
I have added your quick test as part of DPF tests, so it will be easy to check if things still work or not.

PR #75, opened 2018-09-29
Tweak behaviour of NanoWidget as subwidget of another NanoWidget

pdesaulniers:
This would fix the weird behavior I've described in https://github.com/DISTRHO/DPF/issues/68. 

I really don't know if this is a proper fix, but it works with Wolf Shaper and the rectangles example I've shown in the issue. 

It makes the subwidgets pass through this OpenGL code, which sets the correct viewport: https://github.com/DISTRHO/DPF/blob/6d35b690e5399ceaf52b42443b0b3a94587122ac/dgl/src/WidgetPrivateData.hpp#L81-L88

So, calling `rect(0, 0, getWidth(), getHeight());` should draw a rectangle equal to the widget's viewport.

Also, in onNanoDisplay(), you can freely use NanoVG's translate(), reset(), etc. It doesn't seem to affect the OpenGL viewport.

### DIFF
--- a/dgl/src/NanoVG.cpp
+++ b/dgl/src/NanoVG.cpp
@@ -948,7 +948,7 @@ NanoWidget::NanoWidget(NanoWidget* groupWidget)
       nData(new PrivateData(this))
 {
     pData->needsScaling = true;
-    pData->skipDisplay = true;
+    pData->skipDisplay = false;
     groupWidget->nData->subWidgets.push_back(this);
 }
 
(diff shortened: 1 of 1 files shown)


falkTX:
I rather not do this until I am 100% sure it is the right thing.
Sorry it is taking long.

pdesaulniers:
No worries, take your time. I'm not sure this is OK either.

pdesaulniers:
I don't understand why skipDisplay is needed. Why not draw all widgets directly, without special treatment for subwidgets?

In pseudocode:
```
for each widget in window:
    if widget is visible:
        setup the glViewport
        call widget->onDisplay()
```

pdesaulniers:
The pugl-upstream-v2 branch appears to fix my original issue. Closing this PR, as the code is probably not relevant anymore.

PR #79, opened 2018-10-14
Allow negative rotation angle in NanoVG

pdesaulniers:
If I have to rotate a shape counterclockwise by a certain amount of degrees, I would do
```c++
rotate(degToRad(-degrees));
```
This seems to work as expected. I'm pretty sure it's a valid operation.

### DIFF
--- a/dgl/src/NanoVG.cpp
+++ b/dgl/src/NanoVG.cpp
@@ -447,10 +447,8 @@ void NanoVG::translate(float x, float y)
 
 void NanoVG::rotate(float angle)
 {
-    if (fContext == nullptr) return;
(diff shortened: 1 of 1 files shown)


falkTX:
Should the angle be allowed to be 0?
An assertion there would make it possible to catch bad code.

pdesaulniers:
Well, I don't know. This could be annoying for rotating animations, since you'd have to make sure your angle is not zero before rotating.

How about adding this condition to the if?
```c++
if (fContext != nullptr && d_isNotZero(angle))
```

falkTX:
The idea was to catch a bad call, but I see your point about animations.
More power to developers I guess, including hurting themselves :P

PR #80, opened 2018-10-23
Allow disabling logs in Window class

pdesaulniers:
I don't really have a need for these logs most of the time, so I'd rather disable them.


falkTX:
These are just debug things though. I guess you want a debug build without those.

pdesaulniers:
Yes, that's it. Is there a better way of achieving that?

rghvdberg:
please merge this!

falkTX:
I forgot about this being busy with Sonoj.
I propose the following:
1. change the macro to from opt-out to opt-in (ie, always disabled by default)
2. rename the macro DGL_DEBUG_EVENTS
3. remove the stuff in the makefile

pdesaulniers:
Should be good now.

PR #81, opened 2018-10-28
Fix ttl generation for plugins that use events out but not events in

pdesaulniers:
The atom prefix is necessary when the plugin uses events out, so we need to include it: https://github.com/DISTRHO/DPF/blob/master/distrho/src/DistrhoPluginLV2export.cpp#L357

Without the fix in this pull request, the plugin would fail to load in Carla.


falkTX:
Ah yes indeed, that makes sense.
Thanks

PR #83, opened 2018-11-15
Clean up the vestige header

jpcima:
It's especially to get rid of macro definitions in favor of enum. And some minor cleanups.

This is not to address any particular need of DPF itself; it's to benefit other projects which want to use this header, and this repo maintains the most recent revision which I know.

- made constant values available under `enum`
- eliminated check of `WORDS_BIGENDIAN`
- conformity with C language strictest rules

The `enum` change allows the header to be included inside a c++ namespace.
This would allow *another well-known plugin framework*, which removed SDK-less VST support in recent releases, to use vestige as almost compatible drop-in header. (modulo disabling a number unsupported features which this header lacks in comparison with the original)
```
namespace Vst2 {
#include "vestige.h"
}
```

### DIFF
--- a/distrho/src/vestige/vestige.h
+++ b/distrho/src/vestige/vestige.h
@@ -39,146 +39,143 @@
 				( ( (int) c ) << 8 ) |		\
 				( ( (int) d ) << 0 ) )
 
-#define audioMasterAutomate 0
-#define audioMasterVersion 1
-#define audioMasterCurrentId 2
-#define audioMasterIdle 3
-#define audioMasterPinConnected 4
-// unsupported? 5
-#define audioMasterWantMidi 6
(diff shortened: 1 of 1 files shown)


falkTX:
I am against this change actually.
The version I have in DPF matches the one used in Ardour and lv2vst projects (not 100% sure, but I believe so)
If the header is to be changed, I would update it first on those projects.

The SDK is quite tricky business, with Steinberg sending DMCAs to opensource VST2 projects.
So I really do not want to have my own custom version, but just have a copy of someone else's and something that is shared between a few projects.

jpcima:
I understand.. a pity how steinberg likes to make the free software work really difficult.
Closing this for now.

Issue #31, opened 2018-01-27
External UI stays around when closed from the host side


ventosus:
Issue
* The process that the external UI spawns is never killed if the UI is closed via host

Reproduce
* Open Qtractor
* Add ZynAddSubFx plugin
* toggle Plugin Settings->Edit -> opens 1. external UI
* toggle Plugin Settings->Edit -> nothing (should close 1. external UI)
* toggle Plugin Settings->Edit -> opens 2. external UI
* toggle Plugin Settings->Edit -> nothing (should close 2. external UI)

Result
* 2 external UIs stay around, although both should have been closed

Problem
* ExternalWindows::terminateAndWaitforProcess seems not to be called from lv2ui:hide

falkTX:
Which UI type are you using for zyn?

ventosus:
Well, there's only one compiled on ArchLinux, it's a ui:showInterface, aka new external UI.

This is zyn 3.0.3 btw.

`
<http://zynaddsubfx.sourceforge.net#UI>
    lv2:extensionData ui:idleInterface ,
                      ui:showInterface ,
                      <http://kxstudio.sf.net/ns/lv2ext/programs#Interface> ;

    lv2:requiredFeature <http://lv2plug.in/ns/ext/options#options> ,
                        <http://lv2plug.in/ns/ext/urid#map> 
`

falkTX:
I know that.. I'm asking if it's fltk, ntk or zyn-fusion.
Sorry I wasn't clear before.

My guess is that it uses fltk.

ventosus:
Ah you were asking for the toolkit, sorry

It is fltk, indeed

falkTX:
I believe this is now fixed in the `external-ui-tests` branch.
I added an example plugin in order to test this more easily.
Works fine here, would be nice if you could confirm. Thanks!

pdesaulniers:
Yep, seems fixed :)

falkTX:
I was waiting for @ventosus to confirm, since he was the original reporter of the bug.
But seeing the fix in the code directly, I can assume it is really fixed now.
(branch has been merged into develop now)

PR #85, opened 2018-11-25
Exit UI in ~UIVst and ~UiLv2

pdesaulniers:
Fixes https://github.com/DISTRHO/DPF/issues/31 for me in Qtractor. 

I'd like to test this in Carla, but it says: 
```
lv2_rdf_new("http://zynaddsubfx.sourceforge.net") - UI 'http://zynaddsubfx.sourceforge.net#UI' is of unknown type
```
If I change the UI type from UI to X11UI, the UI loads, but the issue remains (UI stays open after hiding) UNLESS I activate 'Run plugins in bridge mode when possible'. It seems like this is because lv2ui_cleanup isn't called on hide, unless bridge mode is activated.

Should I open an issue in Carla?


falkTX:
This is an issue with DPF, not carla I believe.
The explicit UI quit should not be needed. something else is missing...

falkTX:
The UI object was not being deleted for external UIs, leading to this issue.
Now fixed in the `external-ui-tests` branch.
I added an example plugin in order to test this more easily.

See #31
I am closing this PR, as the issue was somewhere else.

PR #86, opened 2018-11-26
Fix some parts of BarBeatTick documentation

pdesaulniers:
I think these errors were copied from the jack docs: http://jackaudio.org/files/docs/html/structjack__position__t.html#a9ba36f6c4eaede3bb1c9b7a85b5f76d1


falkTX:
Yes indeed, the JACK docs should be fixed too.

PR #94, opened 2018-12-26
software rendering with Cairo

jpcima:
I have introduced cairo software rendering in order to make use possible without openGL.

For this, I have modified pugl libraries. Some work was backported from pugl master, not overwriting DPF's own changes, and I wrote some parts myself. 

1. reintroduced conditional backend use in pugl. (by context type and ifdef condition)
2. backported cairo X11.
3. wrote cairo Win32 which didn't exist
4. wrote cairo Quartz for Mac. (upstream pugl has only cairo on GL)

On DPF side, cairo shares most of the dgl code; it is enabled by a definition `HAVE_DCAIRO` as a counterpart for `HAVE_DGL`; currently not accessible from DPF's `Makefile` system.
The drawing methods access a cairo context of the window, by the method `getGraphics`, which only has validity during the drawing period.

There has been light testing on Linux, and also virtualised Windows 7 + SaviHost, Mac 10.6 + Juce AudioPluginHost. These configurations worked.

If cairo support is desired in DPF, I'll port some widgets and publish a clean plugin example.


falkTX:
Wow, this is amazing, was not expecting that :)
It is certainly something desired yes, I gave it a shot once but gave up after upstream also removed (direct) cairo support.
Makes me want to give proper support for DPF, since a lot more people are coming to it now...

There are some changes I would do in DPF UI API though, to make this easier and get some details sorted out too.
For example, the geometry stuff that draws need to be passed a context.
Same for images.
Because of the lack of this "context", I got into issues on Ableton that also uses OpenGL in its UI.
(tricky to properly explain, sorry to not be very clear here)

Your PR looks really clean, congrats on that.
To not make things overly slow right now, I want to merge it, I will just leave a few comments on smal things to change.

dgl/src/Color.cpp
```diff
@@ -105,7 +107,36 @@ Color::Color(const Color& color1, const Color& color2, float u) noexcept
...
+#else
+    float m1, m2;
+    Color col;
+    hue = fmodf(hue, 1.0f);
+    if (hue < 0.0f) hue += 1.0f;
+    fixRange(saturation);
+    fixRange(lightness);
+    m2 = lightness <= 0.5f ? (lightness * (1 + saturation)) : (lightness + saturation - lightness * saturation);
+    m1 = 2 * lightness - m2;
+    auto hue_ = [](float h, float m1, float m2) -> float
```
falkTX: DPF needs c++98 compat in order to build for MacOS 10.6 and PPC.
Please do not use auto in DPF code.

dgl/Window.hpp
```diff
@@ -119,6 +119,10 @@ class Window
     Application& getApp() const noexcept;
     intptr_t getWindowId() const noexcept;
 
+#if defined(HAVE_DCAIRO)
+    cairo_t* getGraphics() const noexcept;
```
falkTX: Can we call this getContext instead of getGraphics? Does it make sense?

dgl/src/Window.cpp
```diff
@@ -203,6 +213,14 @@ struct Window::PrivateData {
             return;
         }
 
+#if defined(HAVE_DGL)
```
falkTX: This part does not look right.
We will have to think of a way to pick either GL or Cairo.
Do we want it to be a compile-time option? or dynamic at runtime?
In case both are available during build time, which ones has priority?
(these are questions for myself too)

falkTX:
There are a bunch of `#if defined(HAVE_DGL)` that I think should be simplified to `#ifdef HAVE_DGL`.

The MacOS part has a lot of duplicated code.
I did not see it initially, I will not merge the PR with it like this.
I wonder how to handle that, probably best to convert the obj-c file into obj-c++ template the class type.
You started a little bit of it, but could be extended a lot more.

falkTX:
ok some something like this:
```
template<class GenericMacViewClass>
class GenericPuglView
{
// regular stuff here
}

class PuglOpenGLView : public GenericPuglView<NSOpenGLView>
{
// opengl specifics here
}

class PuglCairoView : public GenericPuglView<NSView>
{
// cairo specifics here
}
```

I have a feeling you already know of this though, and it is just a matter of making it work with obj-c / cocoa code.
That is the part I did not try yet :P

jpcima:
A thing I observed, somewhat related: `dgl/Makefile` has a reference to the source `Window.mm`, but there exists no file with such name. I forced a build of `Window.cc` as obj-c++ at my side.

For `pugl_osx`, let's try to fix it. I am hesitant to touch openGL because my VM configuration of mac 10.6 can't test openGL at all. It's a old PC and fails with SIGILL in Apple's GL framework..
Someone should recheck openGL to not be broken after the work.

falkTX:
The Window.mm is a fake file in order to trick clang into building the code as obj-c++.
I have an apple mini at home which I use to test this, can do that in a few days once I am back there.

jpcima:
Resolved some code duplication in pugl OSX by moving implementations in common functions.
It's still a repetition of overriding instance methods, but their bodies are at restrained at least to 1 or 2 lines.
I hope it's a fine middle ground solution; it doesn't introduce any sort of low-level trickery.

dgl/Window.hpp
jpcima: I have no particular preference. I renamed it.

dgl/src/Window.cpp
jpcima: I had compile-time choice in mind, but that would sound like a possibility.
I know some users will enjoy a fallback on software rendering when GL support is defective on old machines.

jpcima:
I've made an example plugin available. https://github.com/jpcima/cairo-dpf-example
It illustrates a lack of double buffering which remains to implement for X11 and Windows.
Mac seemed to be fine as it is.

zamaudio:
Nice work!  This is great to see.

jpcima:
>I think it makes sense to typedef the context into something that can be used for GL and Cairo at once.
Then passing the context as a variable of the onDisplay() function.

But it's a breaking change to the interface with existing openGL plugins.

For the geometric shapes, which don't accept currently a context as their argument:
I would introduce `draw(DPF::Context *context)` for both Cairo and GL, and I would let `draw()` still exist but in the case of GL only.

>Regarding MacOS, I would still rather have the templated class, I will take a chance at it in some days if you do not mind.

I think you miss a part of the problem, and the pseudo-code is based on C++ which is not helpful in a Objective-C class model.

Ideally we want the common methods of Cairo and GL to override NSView while residing in a common place. It's a class model like this:
puglclass

But this class model is impossible, because it's multiple virtual inheritance; Objective-C does not support that. The nearest thing I believe you can hack together is a kind of "mixin". Then, it may be a thing like this:
puglmix

Traditional Objective-C doesn't have it, but a small external [library](https://github.com/vl4dimir/ObjectiveMixin) can allow to somewhat accomplish this.

falkTX:
> > I think it makes sense to typedef the context into something that can be used for GL and Cairo at once.
> > Then passing the context as a variable of the onDisplay() function.
> 
> But it's a breaking change to the interface with existing openGL plugins.

Correct, so we would need to keep the old function with a compile warning for some time.
But for now we skip it, as we are just investigating how to properly handle this

> For the geometric shapes, which don't accept currently a context as their argument:
> I would introduce `draw(DPF::Context *context)` for both Cairo and GL, and I would let `draw()` still exist but in the case of GL only.

If you make the context have a default value (nullptr), it will keep backwards compatibility.

I do not know yet what variables need to be part of the context for GL (likely window handle from pugl, so we can call glx stuff).
I suggest have the DPF Context type as this:
```
struct Context {
PuglContextType type; // GL, Cairo, etc perhaps include NanoVG here?
#ifdef HAVE_DCAIRO
cairo_t* cairo;
#endif
};
```

> > Regarding MacOS, I would still rather have the templated class, I will take a chance at it in some days if you do not mind.
> 
> I think you miss a part of the problem, and the pseudo-code is based on C++ which is not helpful in a Objective-C class model.

I think we have a misunderstanding here.
The tree graph I see from you does not match what I have in mind.

I do not want to use Mixins, that seems like a bad idea.
When I am back home, I will give this a try...

falkTX:
I thought more about this and made some decisions...

1. in order to not have this as unmerged PR for too long, I created a develop branch. This will break some things and requires extra work on top (which I will do, no worries) so please rebase or recreate the PR against the develop branch, thank you
there will be conflicts, because I pushed some changes regarding external-ui there. let me know if you have issues there.
The develop branch will serve as the testing ground for it, while developing other things it needs in parallel. which brings me to the next point

2. I will make `dgl.a` build a cairo and opengl versions (if the current system can do so).
I do not want to have cairo as a regular dependency, or have UIs link or using cairo code if they do not need it. if we build 2 separate libraries, we have more control and safety on what we can do.
we can also break backwards compatibility, as the cairo version is a brand new one.
though I'd like to keep things in sync, we have to see how much is possible.

3. I will be doing some resize-ui work on that develop branch as well.

jpcima:
I have rebased on `develop`.
Next, how best to convert the `Image` class which is somewhat tied to the openGL texture formats enums?

GL and Cairo variants of DGL will build to different libraries, so am I correct to assume this eliminates the idea of runtime choice?

falkTX:
Is there a reason for wanting a Cairo + GL library build at the same time?
A plugin UI can only use one or the other (defined during build via `DistrhoPluginInfo.h`), and the libraries are static, so I do not see which use case we might have to build both together..

So this will eliminate run-time choice, correct.

Thanks for the rebase, when I have some time to deal with the makefiles I will merge this

falkTX:
Actually, I meant before to ask to change what the PR is going to be merged against.
I do not know how to call this. In any case, I can manually merge the branch via CLI to handle this.
Sorry for the confusion

PR #96, opened 2018-12-27
LV2 UI file selector

falkTX:
PR for UI file selector, currently targeting LV2.
Using a WIP LV2 spec.

I have it as PR in order to allow comments and review.

WIP, not finalized


falkTX:
similar approach, but working, pushed to develop

PR #97, opened 2019-01-04
software rendering with Cairo (develop)

jpcima:
This is the cairo work, rebased on `develop`


falkTX:
@jpcima I need time to decide on a couple of things, dont consider things finalized until it gets merged to master.

PR #98, opened 2019-01-05
remove useles parentheses in macros - recent compilers don't like them

schnitzeltony:
In case other compilers than gcc require parentheses please ignore this patch.


schnitzeltony:
Ahh this part is copied - was not aware.

I will prepare a patch and from the response we'll see why they are there. If there is a reason we don't see currently I will give note.

PR #99, opened 2019-01-07
Cairo (develop)

jpcima:
Replace the license of the cairo example with ISC. Update file headers.


falkTX:
Thanks for the give away!

PR #100, opened 2019-01-09
Cairo

jpcima:
Resolve type name conflicts on Mac. (`Point` and `DGL::Point`, `Size` and `DGL::Size`)


Makefile.base.mk
```diff
@@ -216,8 +216,15 @@ ifeq ($(HAVE_CAIRO),true)
...
 DGL_FLAGS   += -DHAVE_CAIRO
 
+ifneq ($(WINDOWS),true)
 CAIRO_FLAGS  = $(shell pkg-config --cflags cairo)
 CAIRO_LIBS   = $(shell pkg-config --libs cairo)
+else
+# Always build statically on windows
+CAIRO_FLAGS  = $(shell pkg-config --static --cflags cairo)
+CAIRO_LIBS   = $(shell pkg-config --static --libs cairo)
+CAIRO_LIBS   += -lgdi32
```
falkTX: This is not supposed to be here, gdi32 is already used in `DGL_LIBS`.
If something fails to build, I missed adding that variable..

dgl/src/Window.cpp
```diff
@@ -19,6 +19,10 @@
 
 #include "../Base.hpp"
 
+#if defined(DISTRHO_OS_MAC)
+#import <Cocoa/Cocoa.h>
```
falkTX: This part is not right, I messed up a previous change in pugl_osx.m where Cocoa is imported too late.
The fix should be done in the pugl files.

distrho/src/DistrhoUI.cpp
```diff
@@ -161,6 +161,9 @@ void UI::uiReshape(uint width, uint height)
     glViewport(0, 0, static_cast<GLsizei>(width), static_cast<GLsizei>(height));
     glMatrixMode(GL_MODELVIEW);
     glLoadIdentity();
+#else
+    (void)width;
```
falkTX: To follow the style of the rest of DPF code, this needs:
```
return;

// unused
(void)width;
...
```

examples/ExternalUI/Makefile
```diff
@@ -23,7 +23,9 @@ FILES_UI  = \
 
 include ../../Makefile.plugins.mk
 
+ifeq ($(LINUX),true)
```
falkTX: This check is wrong, there are other systems other than Linux that need `-ldl` like MacOS and Hurd.
So we need a blacklist rather than whitelist.
At least WINDOWS and HAIKU do not use `-ldl`, I think (but not sure) it is the same for BSD

jpcima: It'll be fine to remove this change for now.
The CI attempted a build at first of Windows ExternalUI, because of the naming error in the Makefile.

Makefile.base.mk
jpcima: It's because the link step should place the system libs after the cairo libs on the linker line.
Unfortunately, it's `-lgdi32` which gets listed first, and so this fails.

distrho/src/DistrhoUI.cpp
jpcima: Is `return` mandatory? I find some other sources which use `return`, some which do not.
I can't help but feel upset when one conditional branch has `return`, and the other not.

Makefile.base.mk
falkTX: If it needs `-lgdi32`, something is wrong then. Is it used by cairo somewhere?

jpcima: GDI is used by cairo when win32 surfaces are built in.
It's kinda a cairo-side issue, because `cairo-win32.pc` does not indicate GDI in the library array; so it's necessary that I mention it in a static link setting.

Unless it is linked after cairo, you will observe a series of `undefined references` from `libcairo.a`.

I believe it will be fixed if system libs are collected in an intermediate variable, eg.
```
DGL_LIBS  += $(CAIRO_LIBS)
DGL_LIBS  += $(DGL_SYSTEM_LIBS)
```

falkTX: I see.. I was afraid that was the issue.
I will think of something

dgl/src/Window.cpp
jpcima: In the pugl files it's too late, because the header sequence `OpenGL.hpp` → `ImageBase.hpp` → `Geometry.hpp` has been included at this point, and the types `Point` and `Size` will conflict.

Makefile.base.mk
jpcima: Implemented a solution in 2cf5b89

dgl/src/Window.cpp
falkTX: You're right, I forgot this..
I added something in develop branch that should fix this.
The DGL namespace is not "exported" when building dgl, only for the plugins. So the conflict should not happen anymore.

jpcima: Rebased on latest `develop` branch

PR #101, opened 2019-01-09
Fix some warnings as well as a bug in DSSI

jpcima:
Some warnings fixed
`-Wclass-memaccess`
`-Wparentheses`
`-Winconsistent-missing-override`
`-Wmemsize-comparison`

About the latter, it is a bug detected in DSSI about a misplaced parenthesis.
It will cause the comparison of prefix to work incorrectly.

### DIFF
--- a/distrho/DistrhoPlugin.hpp
+++ b/distrho/DistrhoPlugin.hpp
@@ -590,6 +590,22 @@ struct TimePosition {
               beatType(0.0f),
               ticksPerBeat(0.0),
               beatsPerMinute(0.0) {}
+
+       /**
+          Reinitialize this position using the default null initialization.
+        */
+        void clear() noexcept
+        {
+            valid = false;
+            bar = 0;
+            beat = 0;
+            tick = 0;
+            barStartTick = 0.0;
+            beatsPerBar = 0.0f;
+            beatType = 0.0f;
+            ticksPerBeat = 0.0;
+            beatsPerMinute = 0.0;
+        }
     } bbt;
 
    /**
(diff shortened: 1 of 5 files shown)


distrho/src/DistrhoPluginLV2.cpp
```diff
@@ -169,7 +169,7 @@ class PluginLv2
     void lv2_activate()
     {
 #if DISTRHO_PLUGIN_WANT_TIMEPOS
-        std::memset(&fTimePosition, 0, sizeof(TimePosition));
+        fTimePosition = TimePosition();
```
falkTX: I really dislike this styling..
I would preferably add a `reset` function on the TimePosition struct.

jpcima: How about `clear` as a name instead?
(smart pointers use `reset()`, and can make this a source of silly mistakes)

distrho/DistrhoPlugin.hpp
```diff
@@ -581,24 +581,44 @@ struct TimePosition {
...
     } bbt;
 
    /**
       Default constructor for a time position.
     */
     TimePosition() noexcept
-        : playing(false),
-          frame(0),
-          bbt() {}
+    {
```
falkTX: it is a bit duplicated code, but the values should be initialized in the constructor, it is good practice and we can get warnings if we don't do that.

I am not a native English speaker, I am not sure from clear vs reset which one makes more sense..
So we can have it as "clear".

falkTX:
Thanks!

Issue #102, opened 2019-01-09
Use a different pkg-config program


jpcima:
Hi. Please allow the location of the `pkg-config` program to be user-defined.
eg. permit such a command as `make PKG_CONFIG=i686-w64-mingw32-pkg-config`
There exist several cross-compiling environments for Windows and Mac which provide their pkg-config wrappers.

osch:
Hi, I prepared a simple patch (#103) for providing such a command as `make PKG_CONFIG=i686-w64-mingw32-pkg-config`.

falkTX:
Fixed in develop branch.
Note for mingw, you likely will need to use something like:
```
export PKG_CONFIG="i686-w64-mingw32-pkg-config --static"
```

PR #104, opened 2019-01-10
Makefiles

osch:
For issue #102. Now with base develop-Branch.

### DIFF
--- a/Makefile.plugins.mk
+++ b/Makefile.plugins.mk
@@ -143,12 +143,12 @@ $(BUILD_DIR)/DistrhoUIMain_%.cpp.o: $(DPF_PATH)/distrho/DistrhoUIMain.cpp
 $(BUILD_DIR)/DistrhoPluginMain_JACK.cpp.o: $(DPF_PATH)/distrho/DistrhoPluginMain.cpp
 	-@mkdir -p $(BUILD_DIR)
 	@echo "Compiling DistrhoPluginMain.cpp (JACK)"
-	@$(CXX) $< $(BUILD_CXX_FLAGS) $(shell pkg-config --cflags jack) -DDISTRHO_PLUGIN_TARGET_JACK -c -o $@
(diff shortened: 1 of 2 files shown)


Makefile.plugins.mk
```diff
@@ -6,20 +6,21 @@
 
 # NOTE: NAME, FILES_DSP and FILES_UI must have been defined before including this file!
 
+ROOT_DIR ?= ../..
```
falkTX: This seems completely wrong.
`?=` means environment variables can override this value, this is not what we want at all...

falkTX:
The commit about pkg-config is great, good addition.
I disagree completely on how the other one was done.
Can you do a PR with just the pkg-config stuff? So I can merge that, and we can leave the rest for discussion.

Makefile.plugins.mk
osch: No problem, this was only a very minor enhancement to simplify the usage of `Makefile.plugins.mk` for plugin specific makefiles in another directory structure. To prevent environment variables from overriding this value we could use `ROOT_DIR = ../..` instead and the plugin specific Makefile has to use `override ROOT_DIR = ....`. 

Without this enhancement the plugin specific Makefile has to use 
```
override DPF_PATH = ....
override TARGET_DIR = .....
override BUILD_DIR  = ....
```
IMHO that's also OK so the benefit of introducing `ROOT_DIR` can be seen as negligible.

Makefile.plugins.mk
```diff
@@ -6,6 +6,7 @@
 
 # NOTE: NAME, FILES_DSP and FILES_UI must have been defined before including this file!
 
+PKG_CONFIG ?= pkg-config
```
falkTX: This line is not needed, this file always has `Makefile.base.mk` imported first

falkTX:
Thanks!

PR #105, opened 2019-01-11
Cairo image and widgets

jpcima:
Implemented the drawing methods for geometric shapes on Cairo.
The example remains to write in order to test these.

There are some conditionals on `DGL_OPENGL` to discard some specifics to openGL.
For example, passing a graphics context is mandatory or facultative depending on case.
For the specific case of `Circle`, it's not needed for Cairo to have cache data or a segment count.

I also rewrote the CairoUI example in a coding style more consistent with the rest of DPF.


dgl/src/pugl/.dir-locals.el
```diff
@@ -0,0 +1,9 @@
+((c++-mode . ((indent-tabs-mode . t)
```
falkTX: what kinda file is this?
I guess added by mistake

falkTX:
I really appreciate the effort, but I would do this in a different way.
Instead of changing the function arguments for a bunch of calls, I really think it is better to just add a new constructor to pass the graphics context there.

I want to make `Geometry.hpp` changes in a way that it does not use any `DGL_CAIRO/OPENGL` macros.

falkTX:
I realize this will need a new class variable, probably stored as a const reference.
The "old" constructor without a graphics context argument should be marked as deprecated.
(the actual implementation will use a null context, for which we will have to check for during draw calls)

examples/CairoUI/CairoExampleUI.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+    void parameterChanged(uint32_t index, float value)
+    {
+        // unused
+        (void)index;
+        (void)value;
+    }
+
+private:
+    std::unique_ptr<DemoWidgetClickable> fWidgetClickable;
```
falkTX: I am not sure if `std::unique_ptr` works for c++98, there is a DPF provided ScopedPointer for things like this.

jpcima:
>Instead of changing the function arguments for a bunch of calls, I really think it is better to just add a new constructor to pass the graphics context there.

I dislike quite strongly the idea of having a geometric object, for instance `Rectangle`, reference a graphics object in every instance.
I think a rectangle is best not being more that a point and a size, it's what are `QRect`, and then `QLine` and the like. In DGL, I observe some cases, although not many, which instantiate `Rectangle` for uses other than graphical.

There, carrying the graphics context is never useful. Even then I don't think it makes ontological sense.
What must obligate a shape to relate one particular context, and not any of them?
I think Qt has right idea, when a `QPainter` draws the item, and not the item which says where it draws.

falkTX:
I see, good points.
Maybe GraphicsContext could be a class that can draw the basic shapes..

What do you think of splitting the logic of the geometry classes into new ones, Like DrawableLine, DrawableCircle, etc?

jpcima:
>What do you think of splitting the logic of the geometry classes into new ones, Like DrawableLine, DrawableCircle, etc?

For my personal  thoughts on the matter, I would not diverge much or add to the current model.
However I am tempted to rework a number of things. As written, the preprocessor conditionals are particularly inelegant..

- passing `GraphicsContext&` by reference instead of pointer, to prevent nulls in general case
- enabling no-context methods in openGL case, marked with deprecation attribute.

```
    void draw(const GraphicsContext& context);

#ifdef DGL_OPENGL
    DISTRHO_DEPRECATED void draw();  // implement as `draw(GraphicsContext());`
#endif
```

- if the signature `draw()` must absolutely be had, the cairo context can be available as a thread-local or global. This is what openGL does after all, although not a great idea probably.

- not much to do to about the conditionals in `Circle`, which exclude some details relevant to GL only.
Maybe arrange the constructors in such way that there exists one common constructor, and then the GL-specific constructor.

```
    Circle(const T& x, const T& y, const float size);

#ifdef DGL_OPENGL
    Circle(const T& x, const T& y, const float size, const uint numSegments);
#endif
```

jpcima:
Ok. I agree with your views in good part.
As you say, the geometry classes are not so useful for drawing in plugins, and I agree also.
Especially in cairo, the drawing code is simpler than openGL by a large amount.

As such, it's not so needed to have them. The actual important things would be: point, size, line, rectangle; and we don't care about their drawing.
If a real portable 2d drawing ability were to be implemented, I think frankly it would be better as a nicely-featured painter API.

I think of skipping this and going straight to implementing Image and ImageWidgets, which are an actual important part. Unlike in openGL, I will happily implement `ImageWidgets` without use of `draw()` calls from `Geometry`.

About `Image`, there exist also some elements related strongly to openGL, and Cairo will also have a difficulty replicating the openGL operation on embedded pictures. (in Cairo, it supports ARGB mode not BGRA, and textures will be read in host endian, leading to a compatibility problem on PPC machines).

jpcima:
I'm trying to figure out how best to implement the `Image` class with use of cairo.

In this case, the functionality here was already split as `Image` and `ImageBase`, and it will be fine for cairo to have its own entire implementation of class `Image`.

The problem is how to deal nicely with interacting with image data.
With `png2rgba` being the current official way to embed images:

- the somewhat compatible format is `CAIRO_FORMAT_ARGB32`. This will work on little-endian, fail otherwise. Because the pixel data is taken as native `uint32`, big-endian interprets R/A and G/B swapped, and it's not portable. (it's unlike GL where the byte ordering is flexible)
If the storage order is not matched by system endianness, a byte-swapped buffer must be computed and kept: then memory cost will be 1× footprint in the built file, 2× runtime footprint.

- generate both ARGB and RGBA pixels, and allow compile-time to select which to use.
  This is 1× file footprint, 1× runtime footprint, and complication in the process.

- embed as png data.
It's `k`× file size footprint with data compression, `(1+k)`× runtime footprint.

I'm tempted to go with a png loading option only for simplicity, but I would like your opinion about this.

jpcima:
I'm adding some new work on the cairo branch.
Before anything, I have retracted the geometry methods. Under cairo, they do not provide much of value.

I added support for images, as well as image widgets.
 - cairo loads embedded PNG data. It's the method of least effort, it's portable, it has a moderate footprint on the binary. I provided the artwork generator as `utils/png2c.py`.
- there exist 2 implementations of the `Image` class known as `ImageOpenGL`, `ImageCairo`. `Image` is provided as a typedef on one of these types.
- the `ImageCairo` keeps a cairo surface handle, which associated to an internal reference counter. this counter implements the RAII and sharing.
- the general drawing methods accept `GraphicsContext` as a parameter. for openGL only, the previous signature is still accepted.
- for `ImageKnob`, it will cache the surface which is drawn once, on a change of angle or value. It it's not rotated, it will be a shallow reference to a sub-region of the larger picture.
- I have applied a translation on the origin every time before drawing a widget. It's an equivalent of how the GL viewport is modified, to make widgets draw in local coordinates. This way, the method `image.draw()` works in the intended way.
- I have updated the example, (1) for usage of the local coordinate system (2) for a demonstration of `ImageKnob` and `ImageButton`.

falkTX:
Sorry to take long to get back to you on this.
I do not like all the changes here to be honest.
It is better to split things so I can merge the more "unquestionable" things.

Adding `GraphicsContext` as parameter of draw() looks specially ugly.
I am still with the opinion that it should go on the constructor.

It is fine if you do not want to split things, I can manually merge a few things already, if you do not mind of course..

jpcima:
>It is better to split things so I can merge the more "unquestionable" things.
>It is fine if you do not want to split things, I can manually merge a few things already, if you do not mind of course..

Yes, I think the same. I can rework this into commits, it's not a problem to do it.
I was lazy to reorgarnize it in the first place, after that I managed to put together all pieces for testing.

>Adding `GraphicsContext` as parameter of draw() looks specially ugly.
I am still with the opinion that it should go on the constructor.

This model wants to set in stone that an image can only associate with a particular graphics context.
In cairo, as opposed to GL, the image is not dependent on the window or the backend. Images can draw anywhere.

For example in Juce framework, one can render a widget onto a picture.
I have once implemented this technique as a way to render complex contents onto a tooltip.
There are possibly cases where this might be useful, it's just my humble opinion on such a thing.

dgl/src/WidgetPrivateData.cpp
```diff
@@ -73,6 +76,14 @@ void Widget::PrivateData::display(const uint width,
     }
 #endif
 
+#ifdef DGL_CAIRO
+    cairo_t* cr = parent.getGraphicsContext().cairo;
+    cairo_matrix_t matrix;
+    cairo_get_matrix(cr, &matrix);
+    cairo_translate(cr, absolutePos.getX(), absolutePos.getY());
+    // TODO: scaling with cairo
```
falkTX: cropping is also missing by the way

falkTX:
FYI I merged some of your commits into develop.

I understand the reasons for wanting the split of images and drawing backend, but it makes things tricky to handle for opengl where textures have to be initialized and destroyed.
Deleting the textures needs to be done on the right context, which sometimes gets messy...
(For now I actually disable image texture deletion on OSX as otherwise we get crashes under Ableton Live)
Still need to see how to best solve that, but I want to have things in DPF as simple as possible.
Making an image bound to a graphics context seems to be the easiest path, and one that does not imply a lot of pain... Juce and other frameworks are there for more complex stuff, which is completely fine with me.

Also, I really want to avoid having macro checks (DGL_CAIRO/OPENGL) inside the DGL base header files like Image.hpp.
Having those will make it really tricky for people that do not use DPF Makefiles to build their plugins, as they will need to know additional DPF/DGL macros that need to be defined.

I know it is not something you agree with, but I ask you to change the code to make GraphicsContext be part of Image class constructor.
Thanks for understanding.

jpcima:
Ok I give up on the controversial topic.. done.
One late thought about things: the program `png2c` should rather be `res2c` as it can serve to embed any kind of file.

falkTX:
Thanks! renaming the `png2c` makes sense yeah
I just want to sort out some details on the osx code to make it not so big diff compared to the original code, and then I will merge this.
Need a few more days to sort it out..

jpcima:
ping
I would like to complete the small work remaining if I can, which is the scaling ability.
I don't have a clear idea how it works though, so that I could test it. Do you have some information on this?

falkTX:
help is very appreciated.
I do not know much about cairo scaling, but regarding DPF you can now see an example using the "Info" and the "Meters" example. Info is more interesting, since it uses an internal value to calculate scaling.
But these use nanovg, not raw openGL, so perhaps not that useful for you.

hmm what exactly do you want to know?

jpcima:
I observed the method `Widget::PrivateData::display` has a scaling parameter of type double, and I don't know too much what to do of it.
Should cairo ignore it? Currently I just left there some comments and a todo mention.

falkTX:
All changes have been incorporated into pugl-upstream-v2 branch now.
Only skipped the image knob rotation.

A bunch of stuff was coded in differently, but confirmed to have things working.

Memory management around the image class is not quite right, but will deal with that later.

PR #107, opened 2019-01-12
Adds YK chorus

trebmuh:


falkTX:
Still a bit experimental from what I gathered, but why not add it.
Thanks

PR #108, opened 2019-01-12
Cairo example

jpcima:
Just the basic rework of the cairo example which has been already reviewed.
No use of geometry objects.


falkTX:
this is nice, thanks for the PR split and continued work (and not giving up out right)

PR #116, opened 2019-01-26
AU fixes

zamaudio:
This should make a stereo plugin work on a stereo channel, and a mono plugin to work on a mono channel.
I tried to fix parameter setting, but it still doesn't work.

### DIFF
--- a/Makefile.plugins.mk
+++ b/Makefile.plugins.mk
@@ -76,7 +76,7 @@ AU_BUILD_FLAGS = \
     -DTARGET_OS_MAC
 
 AU_LINK_FLAGS = \
-     -bundle -arch x86_64 -arch i386 \
+     -bundle \
 	 -framework AudioToolbox \
 	 -framework AudioUnit \
 	 -framework CoreAudio \
--- a/Makefile.plugins.mk
+++ b/Makefile.plugins.mk
@@ -72,12 +72,15 @@ AU_BUILD_FLAGS = \
     -Wno-deprecated-declarations \
     -Wno-four-char-constants \
     -Wno-overloaded-virtual \
-    -Wno-unused-parameter
+    -Wno-unused-parameter \
+    -DTARGET_OS_MAC
 
 AU_LINK_FLAGS = \
      -bundle -arch x86_64 -arch i386 \
 	 -framework AudioToolbox \
 	 -framework AudioUnit \
+	 -framework CoreAudio \
+	 -framework CoreServices \
      -exported_symbols_list $(DPF_PATH)/distrho/src/DistrhoPluginAU.exp
 
 # not needed yet
@@ -86,8 +89,6 @@ AU_LINK_FLAGS = \
 #	 -I$(DPF_PATH)/distrho/src/CoreAudio106/AudioUnits/AUPublic/AUViewBase
 #	 -I$(DPF_PATH)/distrho/src/CoreAudio106/AudioUnits/AUPublic/OtherBases
 
-#	 -framework CoreAudio
-#	 -framework CoreServices
 
 # ---------------------------------------------------------------------------------------------------------------------
 # Handle UI stuff, disable UI support automatically
--- a/distrho/src/DistrhoPluginAU.cpp
+++ b/distrho/src/DistrhoPluginAU.cpp
@@ -52,6 +52,8 @@ public:
     {
         CreateElements();
 
+        maxch = GetNumberOfChannels();
+
         AUElement* const globals = Globals();
         DISTRHO_SAFE_ASSERT_RETURN(globals != nullptr,);
 
@@ -68,11 +70,6 @@ public:
     {
     }
 
-    AUKernelBase* NewKernel() override
-    {
-        return new PluginKernel(this, fPlugin);
-    }
-
     OSStatus GetParameterValueStrings(AudioUnitScope inScope,
                                       AudioUnitParameterID inParameterID,
                                       CFArrayRef* outStrings) override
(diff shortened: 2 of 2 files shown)


distrho/src/DistrhoPluginAU.cpp
```diff
@@ -87,31 +84,26 @@ class PluginAU : public AUEffectBase
     {
         DISTRHO_SAFE_ASSERT_RETURN(inScope == kAudioUnitScope_Global, kAudioUnitErr_InvalidParameter);
 
+        outParameterInfo.flags = kAudioUnitParameterFlag_IsReadable | kAudioUnitParameterFlag_IsWritable;
```
FilipeCSnuk: this is actually needed? I got my parameters to appear (and change values) in renoise with just the writable flag.
did you check what happens with parameter outputs?

distrho/src/DistrhoPluginAU.cpp
```diff
@@ -172,6 +164,12 @@ class PluginAU : public AUEffectBase
         return AUEffectBase::GetProperty (inID, inScope, inElement, outData);
     }
 
+    void SetParameter(AudioUnitParameterID paramID,
+                      AudioUnitParameterValue value) override
+    {
+        fPlugin.setParameterValue(paramID, (float)value);
```
FilipeCSnuk: nice, it was this simple!

distrho/src/DistrhoPluginAU.cpp
```diff
@@ -182,43 +180,33 @@ class PluginAU : public AUEffectBase
...
         }
 
-        void Reset() override
-        {
-        }
+        fPlugin.run((const float **)srcBuffer, (float **)destBuffer, inFramesToProcess);
 
-    private:
-        PluginExporter& fPlugin;
+        ioActionFlags &= ~kAudioUnitRenderAction_OutputIsSilence;
```
FilipeCSnuk: ha, the magic flag!

distrho/src/DistrhoPluginAU.cpp
```diff
@@ -182,43 +180,33 @@ class PluginAU : public AUEffectBase
...
-        // processes one channel of interleaved samples
-        void Process(const Float32* const inSourceP,
-                     Float32* const inDestP,
-                     UInt32 inFramesToProcess,
-                     UInt32 inNumChannels,
-                     bool& ioSilence)
-        {
-            DISTRHO_SAFE_ASSERT_RETURN(inNumChannels == DISTRHO_PLUGIN_NUM_INPUTS,);
+        UInt32 i;
+        float *srcBuffer[maxch];
```
FilipeCSnuk: srcBuffer should be a `const float` pointer, so you avoid doing type conversions during the run() function.
this dynamic channel looks like it will be a little pain to support...

distrho/src/DistrhoPluginAU.cpp
```diff
@@ -87,31 +84,26 @@ class PluginAU : public AUEffectBase
...
 
+        outParameterInfo.flags = kAudioUnitParameterFlag_IsReadable | kAudioUnitParameterFlag_IsWritable;
+
         // Name
         {
             const String& name = fPlugin.getParameterName(inParameterID);
             CFStringRef cfname = CFStringCreateWithCString(kCFAllocatorDefault, name.buffer(), kCFStringEncodingUTF8);
 
-            AUBase::FillInParameterName(outParameterInfo, cfname, true); // FIXME true or false??
+            AUBase::FillInParameterName(outParameterInfo, cfname, false);
```
FilipeCSnuk: did you confirm that the memory from `CFStringCreateWithCString` does not need to be released?
(the 3rd argument is about freeing memory from the string)

FilipeCSnuk:
hm I just realized I commented with my company account :P

Makefile.plugins.mk
```diff
@@ -72,12 +72,15 @@ AU_BUILD_FLAGS = \
     -Wno-deprecated-declarations \
     -Wno-four-char-constants \
     -Wno-overloaded-virtual \
-    -Wno-unused-parameter
+    -Wno-unused-parameter \
+    -DTARGET_OS_MAC
```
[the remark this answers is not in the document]
zamaudio: See https://github.com/DISTRHO/DPF/blob/au/distrho/src/CoreAudio106/AudioUnits/AUPublic/AUBase/AUBase.cpp#L53 and https://github.com/DISTRHO/DPF/blob/au/distrho/src/CoreAudio106/AudioUnits/AUPublic/AUBase/AUBase.cpp#L1402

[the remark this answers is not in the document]
zamaudio: We probably also need to add `TARGET_CPU_X86_64`

Makefile.plugins.mk
```diff
@@ -72,12 +72,15 @@ AU_BUILD_FLAGS = \
...
 
 AU_LINK_FLAGS = \
-     -bundle -arch x86_64 -arch i386 \
+     -bundle \
 	 -framework AudioToolbox \
 	 -framework AudioUnit \
+	 -framework CoreAudio \
```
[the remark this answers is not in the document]
zamaudio: Not strictly needed

distrho/src/DistrhoPluginAU.cpp
zamaudio: Just copied a template plugin.  I think it needs to be both readable and writable so you can change the param.

distrho/src/DistrhoPluginAU.cpp
zamaudio: NB: Parameter changes still dont work testing with 10.6.8

distrho/src/DistrhoPluginAU.cpp
zamaudio: We still need to figure out how to create multiple instances of the plugin for multi-mono.

distrho/src/DistrhoPluginAU.cpp
zamaudio: I was playing with this argument to see if params would work.  I didn't end up fixing it so this change seemed to be a no-op to me.

Makefile.plugins.mk
```diff
@@ -72,12 +72,15 @@ AU_BUILD_FLAGS = \
     -Wno-deprecated-declarations \
     -Wno-four-char-constants \
     -Wno-overloaded-virtual \
-    -Wno-unused-parameter
+    -Wno-unused-parameter \
+    -DTARGET_OS_MAC
 
 AU_LINK_FLAGS = \
-     -bundle -arch x86_64 -arch i386 \
+     -bundle \
```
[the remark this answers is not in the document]
zamaudio: You could use CXXFLAGS in the Rez command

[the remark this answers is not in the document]
falkTX: as you noticed yourself, this is not possible.
my idea is to get ouput from compiler with CXXFLAGS and figure out arches from there.

distrho/src/DistrhoPluginAU.cpp
falkTX: but leaving it as false might lead to memory leaks.
I think only if something crashes we need to set it as false

PR #121, opened 2019-01-29
Au fixes4

zamaudio:


distrho/src/DistrhoPluginAU.cpp
```diff
@@ -221,6 +221,8 @@ class PluginAU : public AUEffectBase
             destBuffer[i] = (float *)outBuffer.mBuffers[i].mData;
         }
 
+        updateSampleRate();
```
falkTX: this is not something that should be done in every process, but I'll handle this case later on

Issue #127, opened 2019-02-07
png2rgba: bad conversion of colormap png files


jpcima:
The script `png2rgba.py` mistakes some PNGs for mono-channel images, and converts them to `GL_LUMINANCE` despite being colored.

It can happen on pictures which have been processed by `optipng`.
This set has pairs of pictures before/after optimizing, allowing to reproduce the problem.
[bug-png2rgba.tar.gz](https://github.com/DISTRHO/DPF/files/2841303/bug-png2rgba.tar.gz)

According to the `file` command, the optimization has transformed the input from RGBA to colormap.
```
buttonOff.opti.png: PNG image data, 50 x 30, 8-bit colormap, non-interlaced
buttonOff.png:      PNG image data, 50 x 30, 8-bit/color RGBA, non-interlaced
buttonOn.opti.png:  PNG image data, 50 x 30, 8-bit colormap, non-interlaced
buttonOn.png:       PNG image data, 50 x 30, 8-bit/color RGBA, non-interlaced
```

PR #128, opened 2019-02-07
png2rgba: Support palette-based PNG files

jpcima:
#127 

This is to support palette-based PNG files.
It converts to an intermediate RGBA picture.


falkTX:
What happens if an image has no transparency?
RGBA implies it does..

jpcima:
Better, instead I'm removing the `mode` parameter entirely.
According to documentation of `Image.convert`:
>If mode is omitted, a mode is chosen so that all information in the image and the palette can be represented without a palette.

falkTX:
Fair enough, thanks for the changes.

PR #129, opened 2019-02-08
Fix a build failure with parallel make

jpcima:
When DPF is built once, and then rebuilt with "-j4", this error may occur.

```
Compiling src/OpenGL.cpp (OpenGL variant)
Creating libdgl-opengl.a
Symlinking libdgl.a
ln: failed to create symbolic link '../build/libdgl.a': File exists
```

I can reproduce it every time with `touch dgl/src/OpenGL.cpp` and then `make -j4`.
For some reason, it never occurs on "-j1"; the symlink rule is not invoked when it's on a single make job.


falkTX:
Makes sense, will merge. Thanks!

PR #131, opened 2019-02-12
Replace png2c with a more generic program

jpcima:
It accepts all file types as inputs, not just png files.


falkTX:
Thanks a lot!

PR #132, opened 2019-03-12
Add support for LV2 UI Touch

x42:


distrho/src/DistrhoUILV2.cpp
```diff
@@ -388,6 +388,8 @@ static LV2UI_Handle lv2ui_instantiate(const LV2UI_Descriptor*, const char* uri,
             uiResize = (const LV2UI_Resize*)features[i]->data;
         else if (std::strcmp(features[i]->URI, LV2_UI__parent) == 0)
             parentId = features[i]->data;
+        else if (std::strcmp(features[i]->URI, LV2_UI__touch) == 0)
+            uiTouch = (LV2UI_Touch*)features[i]->data;
```
falkTX: this breaks constness, should cast to const

PR #135, opened 2019-04-06
Document DISTRHO_PLUGIN_WANT_FULL_STATE

pdesaulniers:
The fact that we can use the full state API to set state values on the DSP side was not obvious to me, so I think it should be mentioned in the documentation.

Feel free to change the wording of this summary.

### DIFF
--- a/distrho/DistrhoInfo.hpp
+++ b/distrho/DistrhoInfo.hpp
@@ -545,7 +545,10 @@ START_NAMESPACE_DISTRHO
 
 /**
    Wherever the plugin implements the full state API.
-   Plugins that implement the full state API may set custom state values on the DSP side, without having to pass a message from UI to DSP.
+   When this macro is enabled, the plugin must implement a new getState(const char* key) function, which the host calls when saving its session/project.
+   This is useful for plugins that have custom internal values not exposed to the host as key-value state pairs or parameters.
+   Most simple effects and synths will not need this.
+   @note this macro is automatically enabled if a plugin has programs and state, as the key-value state pairs need to be updated when the current program changes.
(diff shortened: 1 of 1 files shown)


distrho/DistrhoInfo.hpp
```diff
@@ -543,6 +543,13 @@ START_NAMESPACE_DISTRHO
  */
 #define DISTRHO_PLUGIN_WANT_STATE 1
 
+/**
+   Wherever the plugin implements the full state API.
+   Plugins that implement the full state API may set custom state values on the DSP side, without having to pass a message from UI to DSP.
```
pdesaulniers: Perhaps the last part of the sentence could be changed to 

> ... without having to call `setState()` in the UI.

... or something similar.

falkTX: This needs to explain in more detail why and how it is used.

Writing something quick here...

> Wherever the plugin implements the full state API.
> When this macro is enabled, the plugin must implement a new `getState(const char* key)` function, which the host calls when saving its session/project.
> This is useful for plugins that have custom internal values not exposed to the host as key-value state pairs or parameters.
> Most simple effects and synths will not need this.
> @note this macro is automatically enabled if a plugin has programs and state, as the key-value state pairs need to be updated when the current program changes.

falkTX: If this description makes sense to you, please update the PR.
Otherwise let's keep discussion open.

pdesaulniers: Makes perfect sense to me!

falkTX:
nice

PR #139, opened 2019-04-29
Fix invalid atom size on sendNote (LV2)

pdesaulniers:
Carla would complain about the following on sendNote:
```
Warning: LV2 UI sending atom with invalid size! size: 11, padded-size: 16
```
Assuming that this is not a bug in Carla, this seems like a proper fix? 

It looks like sizeof(LV2_Atom_MidiEvent) would add 1 byte of padding (12 bytes instead of 11 bytes), which would lead to this warning. Adding `__attribute__((packed))` to the struct also makes the warning disappear.

### DIFF
--- a/distrho/src/DistrhoUILV2.cpp
+++ b/distrho/src/DistrhoUILV2.cpp
@@ -286,7 +286,7 @@ protected:
         atomMidiEvent.data[2] = velocity;
 
         // send to DSP side
-        fWriteFunction(fController, eventInPortIndex, sizeof(LV2_Atom_MidiEvent), fEventTransferURID, &atomMidiEvent);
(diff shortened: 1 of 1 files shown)


falkTX:
this seems rather strange.. 
can you try this?
```
sizeof(LV2_Atom) + 3
```

your proposed change is not correct, as the host cant actually safely read those many bytes.

pdesaulniers:
Isn't that identical to my fix though? https://github.com/DISTRHO/DPF/blob/bbd99943b5e884132dd7542cab6fef5cc3033057/distrho/src/lv2/atom-util.h#L47-L52

falkTX:
ah you are right, I was confused with the lv2 padding function, but you are not using that here.

PR #141, opened 2019-04-30
Add SVG support + basic MidiKeyboard example

pdesaulniers:
This PR adds the ability to load and display SVGs in DGL UIs. It also adds an example that demonstrates this feature, along with `sendNote`.

The loading and rastering is done using NanoSVG. For now, I've added an `SVG` class, along with a `loadFromSVG(const SVG& svg)` method in the Image class. Later, there should also be some way to display an SVG as a NanoImage or a Cairo image.

There is something important to consider: right now, SVGs are converted to RGBA arrays. This means that we need to recreate a new image every time we want to change the scale of the image, otherwise it won't look crisp. This is not really compatible with the current 'scaling' feature of DGL, which will make the SVG look blurry after scaling, like a standard image.

As for the MidiKeyboard example, it's a really bare-bones keyboard that sends notes from C4 to C6.


dgl/src/Image.cpp
```diff
@@ -92,6 +92,13 @@ void Image::loadFromMemory(const char* const rawData,
     fIsReady = false;
 }
 
+void Image::loadFromSVG(const SVG& svg) noexcept
+{
+    DISTRHO_SAFE_ASSERT_RETURN(svg.isValid(),)
+
+    loadFromMemory((const char*)svg.getRGBAData(), svg.getSize(), GL_RGBA);
```
pdesaulniers: This is pretty bad, as the RGBA data becomes invalid when the SVG is freed. The caller needs to keep both the SVG and the Image 'alive'.

What should we do instead?

examples/MidiKeyboard/MidiKeyboardExamplePlugin.cpp
```diff
@@ -0,0 +1,132 @@
...
+    {
+        return "MidiKeyboard";
+    }
+
+   /**
+      Get an extensive comment/description about the plugin.
+    */
+    const char* getDescription() const override
+    {
+        return "Plugin that demonstrates how to make a MIDI keyboard widget using DPF.";
```
pdesaulniers: I think this description is a bit clumsy.

examples/MidiKeyboard/KeyboardWidget.hpp
```diff
@@ -0,0 +1,486 @@
+#include "Widget.hpp"
+#include "Image.hpp"
+#include "SVG.hpp"
+#include "resources/MidiKeyboardResources.hpp"
+
+START_NAMESPACE_DISTRHO
+
+class PianoKey
```
pdesaulniers: I think this could easily be a Widget. It's rather similar to an ImageButton.

examples/MidiKeyboard/KeyboardWidget.hpp
```diff
@@ -0,0 +1,486 @@
...
+
+    void setCallback(Callback* callback)
+    {
+        fCallback = callback;
+    }
+
+    /**
+       Draw the piano keys.
+     */
+    void onDisplay() override
```
pdesaulniers: Right now, the white and black keys are handled separately, to simplify z-order handling.

If these were widgets, I think it would've been possible to just instantiate the white keys first, and the black ones last. z-order would've been correct without any further adjustments.

examples/MidiKeyboard/KeyboardWidget.hpp
```diff
@@ -0,0 +1,486 @@
...
+        {
+            fBlackKeys[i].draw();
+        }
+    }
+
+    /**
+       Get the key that is under the specified point.
+       Return nullptr if the point is not hovering any key.
+     */
+    PianoKey* tryGetHoveredKey(const Point<int>& point)
```
pdesaulniers: This could easily be optimized, but this approach has the benefit of being simple I guess...

examples/MidiKeyboard/Makefile
```diff
@@ -0,0 +1,52 @@
...
+
+# --------------------------------------------------------------
+# Files to build
+
+FILES_DSP = \
+	MidiKeyboardExamplePlugin.cpp
+
+FILES_UI = \
+	resources/MidiKeyboardResources.cpp \
+	MidiKeyboardExampleUI.cpp
```
pdesaulniers: The KeyboardWidget should probably be split into hpp/cpp files.

examples/MidiKeyboard/Makefile
```diff
@@ -0,0 +1,52 @@
...
+ifeq ($(HAVE_JACK),true)
+ifeq ($(HAVE_OPENGL),true)
+TARGETS += jack
+endif
+endif
+
+ifeq ($(HAVE_OPENGL),true)
+TARGETS += lv2_sep
+else
+TARGETS += lv2_dsp
```
pdesaulniers: This example doesn't really make any sense without OpenGL support.

examples/MidiKeyboard/MidiKeyboardExampleUI.cpp
```diff
@@ -0,0 +1,114 @@
...
+                   17.f / 255.f,
+                   17.f / 255.f);
+
+      glClear(GL_COLOR_BUFFER_BIT);
+    }
+
+    /**
+       Called when a note is pressed on the piano.
+     */
+    void keyboardKeyPressed(const uint keyIndex) override
```
pdesaulniers: I guess there doesn't need to be a separate callback for both 'pressed' and 'released', but I couldn't find a proper name for that callback...

pdesaulniers (commented):
I've left a few comments in the PR. 

By the way, sorry if this PR is too big. I don't mind splitting it in smaller parts if you want.

dgl/SVG.hpp
```diff
@@ -0,0 +1,76 @@
...
+ * NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL
+ * DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER
+ * IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN
+ * CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+#ifndef DGL_SVG_HPP_INCLUDED
+#define DGL_SVG_HPP_INCLUDED
+
+#include "src/nanosvg/nanosvg.h"
```
falkTX: This include is not really needed

dgl/src/Image.cpp
falkTX: add a parameter to allocate memory. if true, do malloc and copy data.

dgl/src/SVG.cpp
```diff
@@ -0,0 +1,95 @@
...
+        free(fRGBAData);
+        fRGBAData = nullptr;
+    }
+
+    NSVGrasterizer* rasterizer = nsvgCreateRasterizer();
+
+    // nsvgParse modifies the input data, so we must use a temporary buffer
+    const size_t bufferSize = dataSize * 4;
+    
+    char tmpBuffer[bufferSize];
```
falkTX: this might segfault if stack size becomes too big.
as this is not a RT operation, allocating memory seems safer

examples/MidiKeyboard/MidiKeyboardExampleUI.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+// -----------------------------------------------------------------------------------------------------------
+
+class MidiKeyboardExampleUI : public UI,
+                              public KeyboardWidget::Callback
+{
+  public:
+    /* constructor */
+    MidiKeyboardExampleUI()
+        : UI(800, 132),
```
falkTX: I would put the initial width and height as static vars, and then use them in the constructor.
This way we avoid calling `getWidth`, as we already know what it is anyway

pdesaulniers:
@falkTX I've polished a few things, and added some way to play the piano using the computer keyboard (very rudimentary for now, will be improved in another PR).

I guess you can start merging the PR piece by piece. At least, the keyboard widget seems to behave as expected, and I'm not aware of any terrible issue in the code.

pdesaulniers:
Hold on, it seems like I broke SVG scaling. I'll need to investigate this tomorrow.

pdesaulniers:
Nevermind, seems like it's working fine. Scaling would make the keyboard bigger than the UI, and it would end up outside of the UI due to a uint underflow.

dgl/src/Image.cpp
pdesaulniers: It seems to me like this would be a pretty wasteful copy. I guess it could be avoided if there was an object that combined both SVG and Image.

falkTX:
After some thought I decided that it is best to place these sorta widgets into a new repo, like DPF-Widgets or similar in name.
So DPF remains small, but we can have the community bring some useful stuff in there.

No time for this now, just letting you know.
I welcome proposals for the repo name :)

pdesaulniers:
Good. Perhaps it would make sense to create a "DISTRHO-Community" organization, and add the repo there. 

I guess the repo could contain more than just widgets (utility classes, DSP stuff, etc). In that case, the name "DPF-Widgets" would be too specific.

As for the piano widget: even if it ends up in a different repo, it would still benefit from key-codes support in pugl. I'll try to create a PR for this soon.

pdesaulniers:
For demonstration purposes, I have updated the piano widget to use keycodes instead of hard-coded QWERTY notes.

PR #143, opened 2019-05-24
Add docs for editParameter and setParameterValue

naught101:


falkTX:
nice and simple, thanks!

PR #148, opened 2019-06-08
Fix LV2 state string exporting

jpcima:
The LV2 exporter does not preserve the newlines in state strings.

If the string matches the condition to be considered long (> 10 chars), newlines are inserted before and after, and the string is not preserved as desired.


falkTX:
That was intentional as to make the output more readable.
I did not expect it to be an issue, but I guess if one compares saved state byte by byte yeah...

Better to have more compatibility over looking nice (specially on files most will just ignore).
I will merge, thanks.

jpcima:
I have a case where the preset is a math expression written as text.
These newlines have ended displaying in the UI, but not in all cases (due to length condition)

PR #151, opened 2019-07-06
Set UTF8 window title using NetWM hints

jpcima:
Set the title with `_NET_WM_NAME` because `XStoreName` cannot handle UTF-8.


falkTX:
Fair enough, seems logical. Thanks, will apply.

PR #152, opened 2019-07-19
Add simple file selection for Windows

jpcima:
This adds the file dialog for Windows. Tried on 7, XP.

It's not async. There aren't any native dialog API which support it.
Anyway, last I checked, VSTGUI used blocking APIs as well. It's better this than nothing I guess.

Conversion back and forth to UTF-16 is the way to support long paths and Unicode.

### DIFF
--- a/Makefile.base.mk
+++ b/Makefile.base.mk
@@ -213,7 +213,7 @@ DGL_SYSTEM_LIBS  += -framework Cocoa
 endif
 
 ifeq ($(WINDOWS),true)
-DGL_SYSTEM_LIBS  += -lgdi32
+DGL_SYSTEM_LIBS  += -lgdi32 -lcomdlg32
 endif
 
 ifneq ($(HAIKU_OR_MACOS_OR_WINDOWS),true)
--- a/dgl/src/Window.cpp
+++ b/dgl/src/Window.cpp
@@ -1281,6 +1281,48 @@ bool Window::openFileBrowser(const FileBrowserOptions& options)
     // show
 
     return (x_fib_show(pData->xDisplay, pData->xWindow, /*options.width*/0, /*options.height*/0) == 0);
+# elif defined(DISTRHO_OS_WINDOWS)
+    // the old and compatible dialog API
+    OPENFILENAMEW ofn;
+    memset(&ofn, 0, sizeof(ofn));
+
+    ofn.lStructSize = sizeof(ofn);
+    ofn.hwndOwner = pData->hwnd;
+
+    // set initial directory in UTF-16 coding
+    std::vector<WCHAR> startDirW;
+    if (options.startDir)
+    {
+        startDirW.resize(strlen(options.startDir) + 1);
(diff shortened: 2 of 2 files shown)


Makefile.base.mk
```diff
@@ -213,7 +213,7 @@ DGL_SYSTEM_LIBS  += -framework Cocoa
 endif
 
 ifeq ($(WINDOWS),true)
-DGL_SYSTEM_LIBS  += -lgdi32
+DGL_SYSTEM_LIBS  += -lgdi32 -lcomdlg32
```
falkTX: this should be optional, only added if file selector is not disabled

falkTX:
non async is quite terrible, but that is win32 :/

I really dislike that the callback is triggered before the function ends.
I imagine having some code that gets into a pending state after that function ends (if it returns true) and then waits for the callback.

We can simulate async here, also ugly but prepares for maybe one day having something better.
So I suggest the following.
1. add a private char pointer to Window class, win32 only
2. save result from win32 file dialog in this pointer, returning from the function without triggering the callback
3. on the window idle function, check if the char pointer is not null, then trigger the callback in there and free memory (carefully copying the class variable first into a temp one, so we can set that var to null)

this would make the file dialog behaviour a bit more consistent across OSes

Makefile.base.mk
jpcima: There doesn't exist any variable yet, in the makefiles, which defines the file dialog to be enabled or not.
We can let --as-needed just do its job, and comdlg32 will not link.

falkTX:
Looking better, I can already merge this. Though would be better if the function only returns true if `pData->fSelectedFile` gets set to a value. right now it returns true regardless if that succeeds or not

jpcima:
>it returns true regardless if that succeeds or not

It will be consistent with async behavior. It's an indication of whether the dialog has been able to lauch, not that it ended in a positive result.

I imagined that a user implements the logic as the following:
- does the file browser open?
- if it doesn't: do a generic fallback solution (eg. custom widget)

When I return false only as `fSelectedFile` is set, it breaks this kind of logic. (when the user has cancelled the dialog)

But, maybe I have misunderstood the greater picture.

falkTX:
You are quite right, that makes sense, I did not think of that.
I start to wonder if we need to provide a method for developers to know the dialog was cancelled, not sure...

There is only 1 small change I would like to do, but it might need a new method in the String class so I will take care of it myself.
I will merge this already, it is good to go for develop branch.

Thanks!

michaelwillis:
Does this mean that DPF can display a file selector in all three target operating systems, or just Linux and Windows? I couldn't tell if the code can accommodate MacOS.

falkTX:
linux and windows.
and linux still being that ugly thing, but well, it works. I want to have something else for it being tried first, and only if the system cannot handle it we fall back to that.
PRs welcome for osx support.

when I have time to do things myself, I want to finalize cairo things, and then AU support

michaelwillis:
Understood. Maybe I'll try to carve out some time to see if I can get it working on MacOS.

PR #156, opened 2019-07-29
Add keycodes support

pdesaulniers:
This is an initial attempt at adding keycodes support to DPF.

These allow to deal with keyboard keys in a layout-independent manner. It's particularly useful for piano-style widgets, to map keyboard keys to piano notes in the correct order.

This code is based on https://github.com/depp/keycode, especially the  `<os>_tohid.c` files and `keycode.h` in the `src` directory.

Tested with the piano widget on Linux, MacOS and Windows, with a couple of different layouts (Colemak, AZERTY, QWERTY).

The keycodes would be wrong when running the UI in Wine, but they would work fine on a real Windows 10 machine. I'm not sure if this is normal.

Perhaps it would be good to test this feature a little more before merging.


dgl/src/pugl/pugl.h
```diff
@@ -122,6 +122,140 @@ typedef enum {
...
+    PUGL_KC_KP_Decimal = 220,
+    PUGL_KC_LeftControl = 224,
+    PUGL_KC_LeftShift = 225,
+    PUGL_KC_LeftAlt = 226,
+    PUGL_KC_LeftGUI = 227,
+    PUGL_KC_RightControl = 228,
+    PUGL_KC_RightShift = 229,
+    PUGL_KC_RightAlt = 230,
+    PUGL_KC_RightGUI = 231
+} PuglKeyCodes;
```
pdesaulniers: We might want to expose this outside of pugl too. Otherwise, the users will have to include `pugl.h` to use these keycodes.

falkTX:
Looks good, but a few things to do:
1. rename COPYING into COPYING.keycode
2. make the keycode array static
3. make KEYCODE_EVDEV_OFFSET static too, and check if opcode < KEYCODE_EVDEV_OFFSET before doing the -= operation

there are a little oddities regarding spaces vs tabs, but I can fix them later myself

also, exposing the keycode API would be nice yes :)

PR #158, opened 2019-07-30
sofd: skip strcpy in case of same source and destination

jpcima:
Fixes undefined behavior: strcpy source and destination overlap
from x42/sofd#3


falkTX:
Makes sense, thanks!

PR #159, opened 2019-07-30
cairo: define geometry functions, fixes debug link

jpcima:
Using the cairo UI type, I found the following problem can be met:
(possibly a matter of recent compilers)

- only under the case DEBUG=true, the linking stage can end in error when functions are declared and not defined. It's the case of drawing functions from Geometry which don't have implementation.

This problem can be verified on DPF's own CairoUI example.

The patch provides dummy implementation of these functions.


falkTX:
Makes sense. At some point we mark this function as deprecated.
btw, we can use `__func__` as a compiler macro to get the function name

PR #160, opened 2019-07-30
Suggestion: keep symbols by default on debug builds

jpcima:
By default, symbols will be removed which is not intuitive behavior, and defeats the initial purpose of a debug build.
I think it's wise that stripping is skipped by default under the debug configuration.


falkTX:
This PR is not needed, have a look at https://github.com/DISTRHO/DPF/pull/160/files#diff-887c01d47725b31b7111b0f412e78c6bL143
If DEBUG=true, LINK_OPTS is empty, which means it does not do stripping

jpcima:
I 've checked it and you are correct, this was certainly a result of my own mistake. thanks

PR #161, opened 2019-07-31
add file dialog for macOS

jpcima:
This is a Mac version of the file dialog.
It's async, vastly inspired from [puglOpenFileDialog](https://github.com/x42/robtk/blob/master/pugl/pugl_osx.m#L527-L564).
Tried in Juce AudioPluginHost.

Some notes:
- it's guarded against having multiple dialogs at once
- when UI is closed with the file dialog unanswered, it stays. (benign probably)

### DIFF
--- a/dgl/src/Window.cpp
+++ b/dgl/src/Window.cpp
@@ -106,7 +106,8 @@ struct Window::PrivateData {
           mWindow(nullptr),
           mParentWindow(nullptr)
 # ifndef DGL_FILE_BROWSER_DISABLED
-        , fOpenFilePanel(nullptr)
+        , fOpenFilePanel(nullptr),
+          fFilePanelDelegate(nullptr)
 # endif
 #else
           xDisplay(nullptr),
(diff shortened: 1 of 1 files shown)


dgl/src/Window.cpp
```diff
@@ -318,6 +327,13 @@ struct Window::PrivateData {
 #elif defined(DISTRHO_OS_MAC)
         mView   = nullptr;
         mWindow = nullptr;
+# ifndef DGL_FILE_BROWSER_DISABLED
+        if (NSOpenPanel* panel = fOpenFilePanel)
+        {
+            [panel release];
+            fOpenFilePanel = nullptr;
```
falkTX: the null set should be on the line above, otherwise it is pointless to create a local variable for it.

falkTX:
Looks good, but the way it does the callback looks weird to me.
I am not sure if this would build with gcc-4.2 from old osx versions, do you have any idea?
(that callback looks like a lambda function)

I would rather have a non-local callback function/method for it.

jpcima:
It must be that gcc does not support the blocks. (closures)
I'll have to check if I can reformulate it passing an object and selector, it appears there exist old API for this.

falkTX:
Confirmed to work under macOS, and builds with gcc-4.2 as well.
Well done, and thank you!

PS: 
image

michaelwillis:
@jpcima Nice! You beat me to it. I'm glad to see this now supported in all three operating systems.

Issue #153, opened 2019-07-23
[macOS VST] Major problems when plugin UI is re-opened


ARival:
Latest Mojave:
Load up any VST built on macOS for macOS, close the plugin window and attempt to re-open it.

Results -
Ableton Live: Bugs out the entire UI, causes program to hang on exit
Audacity: hard locks
Minihost Modular: Plugin UI becomes a black screen

falkTX:
Fixed on develop branch, but not in the best way.
I will see about the best possible case for this, it is a tricky one

falkTX:
Proper fix in fc96e9fd894df36e45aaacac209ca7078ff0629d

I was able to reproduce the issue, and then fix it.
FLStudio had the same problem as they seem to use OpenGL for the macOS GUI, just like Ableton.

![image](https://user-images.githubusercontent.com/1334853/136098793-ec038209-dfcd-4592-8f45-7763e1abe5d3.png)

PR #162, opened 2019-08-01
Fix macOS crash on editor closed and reopened

jpcima:
On Minihost Modular, the window will not be recreated correctly after
closing. It's because the host does not send the effEditClose opcode.

When effEditOpen is reentered, fVstUI!=nullptr would short-circuit the
initialization of the editor, and continue without an existing window.

----
#153
Tried on Minihost Modular. (crossing fingers this doesn't break anything else.)
@ARival can you check other hosts?

### DIFF
--- a/distrho/src/DistrhoPluginVST.cpp
+++ b/distrho/src/DistrhoPluginVST.cpp
@@ -588,7 +588,8 @@ public:
             return 1;
 
         case effEditOpen:
-            if (fVstUI == nullptr)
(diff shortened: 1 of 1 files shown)


ARival:
Hi there!

All 3 of my test hosts work great! XOXO

falkTX:
There was something I wanted to do on top of this, but completely forgot what it was.
So I will merge, thanks.

jpcima:
Maybe closing events from the window should be listened to on plugin side?

PR #165, opened 2019-08-03
Fix crashes in case of absent openGL

jpcima:
- checks context nullity at entry of a NanoVG function where it's forgotten
- on openGL context initialization after the first one, the function pointers would not be checked for null, resulting in crash


falkTX:
I don't particularly like the way the `needsInit` check is before every function, but I cannot think of anything else that would not require 2 different macros.
So applying this now, thanks!

PR #166, opened 2019-08-03
Resolve all the lv2lint issues

jpcima:
- Since the macro conditional logic began to get intricate, I have factored the code in the exporter such that there needn't be so much care taken regarding the syntax, and reduce error possibilities. ed1572a

It's a starting point so I can add and remove extension data easily, and I verified the examples manifests before/after using diff. Apart from some whitespace, it was identical.

- Licenses were changed to URI. 4cfa3d3
  At the same time, cairoUI's metadata was put in consistency with the rest of examples.

- Missing extension data were added. I checked it against LV2 wrappers source code.
  I found the KX program `Interface` to be mistaken for `UIInterface` and fixed it.

- Implemented the worker's response function. It's not supposed to be null.
  Implementation does nothing.

### DIFF
--- a/distrho/src/DistrhoPluginLV2.cpp
+++ b/distrho/src/DistrhoPluginLV2.cpp
@@ -846,6 +846,14 @@ public:
 
         return LV2_WORKER_SUCCESS;
     }
+
+    LV2_Worker_Status lv2_work_response(uint32_t size, const void* body)
+    {
+        (void)size;
+        (void)body;
+
+        return LV2_WORKER_SUCCESS;
+    }
 #endif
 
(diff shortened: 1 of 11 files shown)


falkTX:
Reviewing this, I like the approach, but I do not like changing the license text to be a URL.
In DPF, the license can be a short string (like "MIT" or "GPLv2") or a full URL.

Not everyone is going to know the proper URL for each license, DPF should make it easier.
In cases of licenses like "MIT" where they are common and known, we can map that to a full URL when exporting LV2 data.
I am going to merge the PR manually, but skipping that change. We can figure out a nice way to handle it in another PR or ticket.

Issue #164, opened 2019-08-02
FR: Add parameter property to set the LV2 port comment property


SpotlightKid:
Some hosts (e.g. Ardour) use this to show a tooltip for the parameter in the generic UI.

![ardour-lv2-port-doc](https://user-images.githubusercontent.com/5719151/62389464-2f814680-b560-11e9-8a42-f85be32000dc.png)

jpcima:
If this the `rdfs:comment` property?
For the `lv2:documentation` property, I don't get a tooltip from Ardour (version 5).

SpotlightKid:
Yes, sorry, you are right, I neglected to check NY wrong assumption.

Host probably should display the documentation property too, but that's a different issue :)

falkTX:
This has been merged now

PR #169, opened 2019-08-07
Add LV2 parameter comments

jpcima:
#164

Ability to set a description on LV2 parameters.

Also, there was a case of extra newlines in the plugin's comment.
I removed these newlines so the plugin text would go into the metadata unchanged.


falkTX:
Makes sense, it is easy to add and useful.
I will merge and then add a mention that is LV2 only.

PR #170, opened 2019-08-07
Fix an 'opts:interface not defined' warning

jpcima:
Fix a remaining lv2 warning. Some change must have been lost in transit during the merging.
The plugin provides the options interface under all conditions; update extension data in order to reflect this.


falkTX:
I think I misread the commit diff before, anyway, thanks.

PR #172, opened 2019-08-13
[VST] Round integer param values after de-normalizing (fixes #171)

### LINKED ISSUE #171: Wrong handling of kParameterIsInteger in PluginVst::vst_setParameter
In `PluginVst:.vst_setParameter` the value of params with the `kParameterIsInteger` hint, which is received by this method being in the range 0.0f - 1.0f, gets rounded, causing it to be either 0 or 1, so the value passed to `fPlugin.setParameterValue` is always either `ranges.min` or `ranges.max`.

I think this bug was introduced by c2b9c40333a7b5e066dcb8d60fa023373b639324.

SpotlightKid:
Has the additional (beneficial) effect, that params, which also have `kParameterIsBoolean` hint, are displayed as 0 and 1, not 0.0 and 1.0.

### DIFF
--- a/distrho/src/DistrhoPluginVST.cpp
+++ b/distrho/src/DistrhoPluginVST.cpp
@@ -925,14 +925,13 @@ public:
         const uint32_t hints(fPlugin.getParameterHints(index));
         const ParameterRanges& ranges(fPlugin.getParameterRanges(index));
 
+        value = ranges.getUnnormalizedValue(value);
+
         if (hints & kParameterIsBoolean)
         {
             const float midRange = ranges.min + (ranges.max - ranges.min) / 2.0f;
-
             value = value > midRange ? ranges.max : ranges.min;
         }
-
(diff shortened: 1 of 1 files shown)


jpcima:
What about the boolean case too? if it's only raw 0-1 values which enter the callback, it seems suspicious for it to be compared in the min-max domain value.
Probably, unnormalizing should precede even the boolean case.

SpotlightKid:
Yes, you're probably right, though in practice `kparameterIsBoolean` params are probably always defined with `ranges.min = 0.0` and `ranges.max = 1.0`, so it makes no difference then.

But in case they're not, yes, the value should be unnormalized before comparing it to the min/max range. I'll update the PR.

falkTX:
This makes sense.
But one thing to note is that it is possible to set custom ranges (outside of 0-1) when using integers in vst. Somehow that is not available for floats, beats me why...

PR #173, opened 2019-08-13
Adds more DPF-based plugins

trebmuh:


falkTX:
I am not that comfortable on adding WIP unfinished plugins, but we can just say they will be ready soon in this case :)

PR #178, opened 2019-08-18
Suppress the copy methods of ParameterEnumerationValues (fixes #177)

### LINKED ISSUE #177: ParameterEnumerationValues not safe to copy
This code below does double-free and crashes.
it should either: 1. forbid copy 2. have copy initializer + assignment

```
    ParameterEnumerationValues a;
    a.values = new ParameterEnumerationValue[1];
    ParameterEnumerationValues b = a;
```

jpcima:
Prevent copying the instances of `ParameterEnumerationValues`.

### DIFF
--- a/distrho/DistrhoPlugin.hpp
+++ b/distrho/DistrhoPlugin.hpp
@@ -365,6 +365,8 @@ struct ParameterEnumerationValues {
             values = nullptr;
         }
     }
+
+    DISTRHO_DECLARE_NON_COPY_STRUCT(ParameterEnumerationValues)
 };
 
 /**
--- a/distrho/src/DistrhoDefines.h
+++ b/distrho/src/DistrhoDefines.h
@@ -128,7 +128,12 @@ private:                                           \
     StructName& operator=(StructName&) = delete;     \
     StructName& operator=(const StructName&) = delete;
 #else
-# define DISTRHO_DECLARE_NON_COPY_STRUCT(StructName)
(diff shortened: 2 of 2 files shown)


falkTX:
The idea is good, but `ParameterEnumerationValues` is a struct, not a class.
So:
1. no need for `private`
2. we should use `DISTRHO_DECLARE_NON_COPY_STRUCT`

jpcima:
I examined the macro `DISTRHO_DECLARE_NON_COPY_STRUCT`.
When c++ version is less than 11, it expands to nothing.

falkTX:
Because it would lead to build errors otherwise.
The "= delete" syntax is a C++11 feature, and structs have no private members, so on old compilers there is nothing we can do..

jpcima:
>structs have no private members
I don't understand your remark. Structures can have private members.
Exactly like it is now, the commit works in C++98.

falkTX:
Oh you are quite right.
I had this wrong misconception of structs not having support for private/protected all this time, wtf! :joy: 
hmmm need to refactor those macros then hah

jpcima:
Ah ouf, glad we are in understanding about this matter. :smile:
I'm going to edit things as necessary.

falkTX:
Thanks!
At a later time we need to get rid of the struct/class separate definition.
Sadly have to keep them around for a bit, due to backwards compatibility, but this is a case where a compiler warning makes sense.

PR #179, opened 2019-08-19
Allow generic units to pass validation

jpcima:
It would fail with this error below.
```
    [FAIL]  Units
              units_unit not a URI or object
              seeAlso: <http://lv2plug.in/ns/extensions/units#unit>
```


falkTX:
Makes sense, thanks!

PR #180, opened 2019-08-23
Fix the build in case the DISTRHO namespace is not used

jpcima:
The build was broken in case the plugin is compiled with `-DDONT_SET_USING_DISTRHO_NAMESPACE`.


falkTX:
Thanks, good catch.
I will add this to CI so we can catch it next time (also custom DPF namespace)

PR #181, opened 2019-08-25
InfoExampleUI.cpp: remove unnecessary assignments

shlyakpavel:


falkTX:
Those lines are intentional, so it is easy to add more stuff below it when the need comes.

shlyakpavel:
Okay

PR #182, opened 2019-08-29
Add the `VERBOSE` option to makefiles

jpcima:
This is a flag `VERBOSE=true` in order to print the compilation commands.
I found myself desiring this option many times in order to inspect the commands, which needs to edit out the @ symbols from the makefiles.


falkTX:
I have tweaked the gcc commands several times until they got just right, but I understand from an outside view wanting to see what is being done.
Does not hurt, will merge, thanks

PR #185, opened 2019-09-27
Fix setting the bypass parameter at init

jpcima:
There is a problem reloading a saved LV2 plugin state having enabled=off (aka bypass=on).

The comparison of Port value vs. Last parameter value is producing `0 != 0` initially, and plugin does not receive setParameterValue for the bypass. It happens in Ardour.

It's not the UI problem reported before, this is yet something else.

For detail see: https://github.com/jpcima/stone-phaser/issues/2
Also: https://forum.moddevices.com/t/stone-phaser-a-vintage-analog-phaser-pedal/

If the proposition is fine as a solution, this would add a pair of methods which invert as needed, when exchanging value between parameter and port, in either direction.

### DIFF
--- a/distrho/src/DistrhoPluginLV2.cpp
+++ b/distrho/src/DistrhoPluginLV2.cpp
@@ -166,6 +166,48 @@ public:
 
     // -------------------------------------------------------------------
 
+    bool getPortControlValue(uint32_t index, float *value) const
+    {
+        const float* control = fPortControls[index];
+
+        if (control == nullptr)
+            return false;
+
+        switch (fPlugin.getParameterDesignation(index))
+        {
+        default:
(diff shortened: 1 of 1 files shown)


falkTX:
designations makes this a bit complicated, thanks for the fixes.
I have some requests though:
1. use c++ references instead of pointers (and put reference operator near the type, not the value)
2. put the `fPortControls[index]` check inside its own `if` block, and make the pointer const
3. `setPortControlValue` return value is never used, so make it void

jpcima:
It's done, seems to run alright in Ardour on my end, I re-ran it a few times.

falkTX:
nice, thanks a lot

Issue #186, opened 2019-10-07
LV2 preset does not export bypass correctly


jpcima:
Hi, I discovered the preset exporter to be affected as well by a bypass issue.
On recalling a preset, this message appears: `error: Preset port 'dpf_bypass' is missing`

As it appears, the exporter needs also to receive the usual treatment.
- exchange `dpf_bypass` with `lv2_enabled`
- invert the default value

falkTX:
Fixed by your own commit, imported as 8616522d5a13398fda032393780f5749d47192fc

PR #188, opened 2019-10-17
Preset bypass

jpcima:
#186

This fixes the way how the bypass parameter is stored in the preset.
- the special symbol `lv2_enabled`
- the value inversion

On loading the preset, jalv would otherwise complain about the missing parameter "Bypass".


falkTX:
Thanks for the bugfix, though I dislike having a bypass parameter where it does not really belong (since it does nothing in the presets example).
It is best to have another example plugin that uses it, something that can achieve a smooth on/off.

Issue #190, opened 2019-10-28
don't hardcode -msse and -msse2 for non-x86 platforms


BrainDamage:
https://github.com/DISTRHO/DPF/blob/4c8d448b10eeeb55f507b6e55e6095e108560fe6/Makefile.base.mk#L109
hardcodes -msse and -msse2 build flags for every architecture, forcing one building DPF on platforms such as arm to manually edit the file instead of just settings CFLAGS to set optimizations like one would espect normally.
Please edit the file so that those two flags are appended just for x86 systems, the rest of the platforms can at least append their own without editing files this way

SpotlightKid:
Isn't this fixed now by #191?

falkTX:
Yes it is :)

I need to do the same for Carla and Cadence projects at some point too

Issue #109, opened 2019-01-16
Re-add optimisation flags for various platforms


zamaudio:
DPF used to have opt flags for various platforms but now seems to be hardcoded to SSE/x86 based flags.  Can we have these added back in based on the platform, eg ARM + neon etc

PR #191, opened 2019-10-30
Improve detection based on GCC triple, detect CPU type

jpcima:
Hi, this adds a detection of common processors based `gcc -dumpmachine` output.

When either type of Intel processor is detected, the flags `-msse -msse2` will be set, and not in any other case.

Also it makes the OS detection a bit more rigorous, which has allowed to find and resolve an issue:
with the triple containing "linux-gnu", both LINUX and HURD are going to be identified.
It was resolved by checking for Hurd using the name "gnu" exactly.

If the OS is not matched correctly, it will raise an error.

#190
also related: #109

### DIFF
--- a/Makefile.base.mk
+++ b/Makefile.base.mk
@@ -135,6 +135,10 @@ ifeq ($(CPU_I386_OR_X86_64),true)
 BASE_OPTS += -msse -msse2
 endif
 
+ifeq ($(CPU_ARM),true)
+BASE_OPTS += -mfpu=neon-vfpv4 -mfloat-abi=hard
+endif
+
 ifeq ($(MACOS),true)
 # MacOS linker flags
 LINK_OPTS  = -fdata-sections -ffunction-sections -Wl,-dead_strip -Wl,-dead_strip_dylibs
--- a/Makefile.base.mk
+++ b/Makefile.base.mk
@@ -11,6 +11,8 @@ CXX ?= g++
 # ---------------------------------------------------------------------------------------------------------------------
 # Auto-detect OS if not defined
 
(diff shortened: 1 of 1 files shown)


falkTX:
I love the idea, but I do not like this PR tbh

First, the triplet CLI is quite ugly and uncompreensible.
If I need to edit it for some reason, I have no idea wtf is going on there.

And second, we should not error out in case of unknown platform, though this one is easily fixed (just remove the error).
DPF will fallback to POSIX and X11 if not running Linux, so it can build on BSD, Hurd and any other. The Linux-specific calls are protected with a compiler macro.

I like the detection for x86.
But it needs to be made in a way that is more understandable, and less black magic

jpcima:
I am modifying this PR to take care of the CPU only. (no Hurd)

The thing of any text handling with Makefile, it's it has to deal with the limited ability of builtin functions, always going to be hackish unless involving external tools.

Considering the former method I used to extract properly the triplet items, I wasn't able to come up with a simpler expression.

falkTX:
well, I fine the current patch quite nice, so we can apply if it you have no more pending changes.
for arm actually, I would add a few flags in there to make sure hardware floats are used.
I recommend `-mfpu=neon-vfpv4 -mfloat-abi=hard` since this is what Pi3 and most boards after it use.

we must not forget that `make NOOPT=true` has always been a thing.

thanks a lot for the effort.

jpcima:
I've just added ARM flags, and as I'm concerned I'm fine about merging this PR now.

falkTX:
Great. merging now then.

Issue #192, opened 2019-10-30
Port groups


jpcima:
It will be interesting to support the Port Groups feature of LV2 in DPF.

- a port group has a name and a symbol;
- a port group is input or output, and additionally may inherit from a predefined class of the specification;
- a parameter or audio port belongs to zero, one (or more?) port group(s).

I have started to think how it is best implemented, and I wondered about this:
may ports belong in multiple groups? does it ever make sense?

Edit: some information relevant to port groups in various plugin formats
https://forum.juce.com/t/plug-in-parameter-groups/29409

falkTX:
A port can only belong to a single group.
It does not have to be input or output, that should not matter. whatever it is in/out depends on which port it gets associated with.

falkTX:
Implemented in 495dcc4f03d6a2ed0d30c56c885e4eb7476f4488.

PR #193, opened 2019-11-01
Port groups proposal

jpcima:
#192

Proposing a first implementation for the port groups feature.
It's left uncommented for now on purpose, I will do later if the design is favorably reviewed.

The `Parameter` example was updated. For user point of view, it's as follows:
1. set a new option `DISTRHO_PLUGIN_WANT_PORT_GROUPS` to `1`
2. set the `group` property on an `AudioPort` or `Parameter` (`0-N` index, or `-1` for the "root" pseudo-group)
3. implement `initPortGroup` to set name and symbol

On plugin-side, will invoke `initPortGroup` for each `0-N` index set at the step 2 above, and collect results into an array, by their respective indices.
It will still work if there were gaps within the `0-N` domain, although it will assert.

In the exporter, the name is mapped to `lv2:name` attributes.
I recall some IRC discussion suggesting that `rdfs:label` is correct instead, so I'm not entirely sure about this point. Regarding Jalv, `lv2:name` is the recognized attribute and none other.


distrho/src/DistrhoPluginInternal.hpp
```diff
@@ -203,6 +222,36 @@ class PluginExporter
         for (uint32_t i=0, count=fData->parameterCount; i < count; ++i)
             fPlugin->initParameter(i, fData->parameters[i]);
 
+#if DISTRHO_PLUGIN_WANT_PORT_GROUPS
+        std::set<int32_t> portGroupIndices;
+
+        for (uint32_t i=0; i < DISTRHO_PLUGIN_NUM_INPUTS+DISTRHO_PLUGIN_NUM_OUTPUTS; ++i)
```
SpotlightKid, line 228: This needs to be (for plugins with no audio inputs/outputs):

```
#if DISTRHO_PLUGIN_NUM_INPUTS > 0 && DISTRHO_PLUGIN_NUM_OUTPUTS > 0
        for (uint32_t i=0; i < DISTRHO_PLUGIN_NUM_INPUTS+DISTRHO_PLUGIN_NUM_OUTPUTS; ++i)
            portGroupIndices.insert(fData->audioPorts[i].group);
#endif
```

distrho/src/DistrhoPluginLV2export.cpp
```diff
@@ -702,6 +732,56 @@ void lv2_generate_ttl(const char* const basename)
...
+                pluginString += "\n<" DISTRHO_PLUGIN_URI "#portGroup_" + pgroup.symbol + ">\n";
+
+                bool isInput = false;
+                bool isOutput = false;
+
+                for (uint32_t i=0; i < DISTRHO_PLUGIN_NUM_INPUTS && !isInput; ++i)
+                    isInput = plugin.getAudioPort(true, i).group == pgroupIndex;
+
+                for (uint32_t i=0; i < DISTRHO_PLUGIN_NUM_OUTPUTS && !isOutput; ++i)
+                    isOutput = plugin.getAudioPort(false, i).group == pgroupIndex;
```
SpotlightKid, line 755: This needs to be (for plugins with no audio inputs or outputs):

```c++
#if DISTRHO_PLUGIN_NUM_INPUTS > 0
                for (uint32_t i=0; i < DISTRHO_PLUGIN_NUM_INPUTS && !isInput; ++i)
                    isInput = plugin.getAudioPort(true, i).group == pgroupIndex;
#endif

#if DISTRHO_PLUGIN_NUM_OUTPUTS > 0
                for (uint32_t i=0; i < DISTRHO_PLUGIN_NUM_OUTPUTS && !isOutput; ++i)
                    isOutput = plugin.getAudioPort(false, i).group == pgroupIndex;
#endif
```

SpotlightKid:
There are a few changes necessary to support plugins with no audio inputs or outputs (e.g. MIDI filters). See my comments on the commit changes.

distrho/src/DistrhoPluginLV2export.cpp
```diff
@@ -324,6 +325,9 @@ void lv2_generate_ttl(const char* const basename)
         pluginString += "@prefix mod:  <http://moddevices.com/ns/mod#> .\n";
 #endif
         pluginString += "@prefix opts: <" LV2_OPTIONS_PREFIX "> .\n";
+#if DISTRHO_PLUGIN_WANT_PORT_GROUPS
+        pluginString += "@prefix pg:   <" LV2_PORT_GROUPS_PREFIX "> .\n";
```
SpotlightKid, line 329: Minor niggle: the indentation of the URLS increase by one space in the current Git version of DPF

jpcima:
Branch is rebased and problems resolved in case of no I/O audio ports. thanks @SpotlightKid

falkTX:
I am considering this now, design-wise seems fine to me, but I would like to do some tweaks.
I do not think it makes sense to introduce a new C macro for this, for example, as there is no new host features required to make use of it. Alike `initAudioPort`, we can have a default implementation (in this case I guess it does nothing).

So, leaving a few notes here as TODO for this feature:
1. Remove use of DISTRHO_PLUGIN_WANT_PORT_GROUPS macro
2. Convert kPortGroupDefault from int32 to uint32 (so type matches the one used in Parameter struct)
3. Rename getPortGroup to getAudioPortGroup
4. Rename kPortGroupDefault to kPortGroupNone
5. Add a few pre-set groups that are commonly used, we can look at lv2 for inspiration
6. Documentation

Additionally, but doesn't have to be part of this PR:
1. VST2 parameter grouping, using effGetParameterProperties and category stuff (needs a host for testing, I will add support for this in carla since I expect most hosts dont use this)
2. JACK meta-data handling for the audio ports

@jpcima do you feel like tackling this, or should I go at it myself?

jpcima:
> do you feel like tackling this, or should I go at it myself?

Not at the moment, please go ahead and do.

falkTX:
Done in 495dcc4f03d6a2ed0d30c56c885e4eb7476f4488 with some base code from this PR, thanks very much!

PR #194, opened 2019-11-03
Fix swapped minor/micro version number when exporting LV2 TTL

SpotlightKid:
Reference: http://lv2plug.in/ns/lv2core/lv2core.html#minorVersion

See also `d_version` in [DistrhoUtils.hpp](distrho/DistrhoUtils.hpp#69).

### DIFF
--- a/distrho/src/DistrhoPluginLV2export.cpp
+++ b/distrho/src/DistrhoPluginLV2export.cpp
@@ -691,15 +691,15 @@ void lv2_generate_ttl(const char* const basename)
             const uint32_t version(plugin.getVersion());
 
             const uint32_t majorVersion = (version & 0xFF0000) >> 16;
-            const uint32_t microVersion = (version & 0x00FF00) >> 8;
-            /* */ uint32_t minorVersion = (version & 0x0000FF) >> 0;
+            /* */ uint32_t minorVersion = (version & 0x00FF00) >> 8;
+            const uint32_t microVersion = (version & 0x0000FF) >> 0;
 
             // NOTE: LV2 ignores 'major' version and says 0 for minor is pre-release/unstable.
             if (majorVersion > 0)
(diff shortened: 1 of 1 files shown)


SpotlightKid:
I'm also not sure about what the `if (majorVersion > 0) minorVersion += 2;` is supposed to solve. But that would be for another PR, I guess.

jpcima:
At first look, I think this fix to make sense, it will be nice to have it.

>I'm also not sure about what the if (majorVersion > 0) minorVersion += 2; is supposed to solve. But that would be for another PR, I guess.

From what I learnt recently on LV2 versioning, a version is considered prerelease under either these conditions:
1. either minor or micro is odd
2. micro is 0

As I understand, this check is to prevent the case 2. to happen.
By DPF's definition, major=0 is intended for prerelease, major>0 for final.

SpotlightKid:
> micro is 0

Actually, "minor == 0":

> An odd minor or micro version, or minor version zero, indicates that the resource is a development version.

http://lv2plug.in/ns/lv2core#minorVersion

> As I understand, this check is to prevent the case 2. to happen.

But why? Why can't one have pre-releases when the major version is >= 1?

falkTX:
is that a dpf question or lv2 question?

SpotlightKid:
A DPF question. In LV2, the major version is the plugin URI, so the major version set with d_version is irrelevant. I don't see anything in the LV2 spec forbidding an odd minor ot micro version, i.e. a pre-release, even after a previous release had even verdion numbers.

falkTX:
To come back to this, my point was to prevent 1.0.0 to turn up as 0.0 and thus be marked as experimental/unstable on lv2 side.

falkTX:
Changes have been pushed to develop branch a while ago, forgot to close this PR, doing that now.

PR #195, opened 2019-11-04
Export LV2 scalePoint for integer ports as integer values not float

SpotlightKid:
Signed-off-by: Christopher Arndt <chris@chrisarndt.de>

### DIFF
--- a/distrho/src/DistrhoPluginLV2export.cpp
+++ b/distrho/src/DistrhoPluginLV2export.cpp
@@ -561,7 +561,11 @@ void lv2_generate_ttl(const char* const basename)
                                 pluginString += "        [\n";
 
                             pluginString += "            rdfs:label  \"\"\"" + enumValue.label + "\"\"\" ;\n";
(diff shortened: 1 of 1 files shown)


falkTX:
This is helpful, thanks.
But I think the value should be rounded. If the value is 2.999, for example, flushing it to 2 doesn't seem right.
We can easily do this by adding 0.5 to the value `int var = (int)(floatvar + 0.5f);` which will truncate/flush-down the value in a rounding way.

SpotlightKid:
Ok, I've pushed a commit that does as you suggest.

falkTX:
Thanks!
I hate to be picky (hmm or maybe not :P) but such variables should be marked `const` :)

SpotlightKid:
Ok, btw, which naming style for local variables? All lowercase, no underscore ok?

falkTX:
CamelCase for class names and variables, with classes started with an uppercase letter.

SpotlightKid:
Feel free to squash the commits.

falkTX:
thanks, will do

Issue #198, opened 2019-11-14
Resizing issue in Reaper Mac


jpcima:
Hi, I have now a possibility to test Reaper on MacOS, so I investigated size issues.

There will be a problem where host and plugin are going to resize each other in a loop, in case the window is set to resizable, and the window is going to shrink or expand indefinitely over a span of some seconds.

A thing to know is that Reaper's containing view is not sized identically to the content.
The Pugl view is offset by some Y inside of the container, this top part is for Reaper to place its own controls.

![rect856](https://user-images.githubusercontent.com/17614485/68872796-e4e69d00-06fe-11ea-9894-0087579c81c3.png)

**Problem**

This code is the curent source of problem.
```
if (view->user_resizable) {
    [impl->view setAutoresizingMask:NSViewWidthSizable|NSViewHeightSizable];
}
```

- when Pugl resizes, hosts wants to resize the parent with some extra height to give room for the controls bar;
- when Host resizes, Pugl wants to let Cocoa do its autoresize, proportionally to parent.

Because both don't get in agreement on their view size, it goes on continually.

**Resolution**

It's possible to implement a custom autoresize, by putting an observer on parent view's frame size.

In Reaper specifically, resizing is able to be resolved by subtracting XY offset from parent's size to determine new target size.
```
NSRect frame = [self frame];
NSRect pframe = [parent frame];
[self setFrameSize:NSMakeSize(pframe.size.width - frame.origin.x, pframe.size.height - frame.origin.y)];
```

However, this method would make an assumption of bottom-right alignment in the parent; I think it will not be correct as general solution.

jpcima:
It remains to do the same job with Cairo.
The solution will need to be a little different. NSView does not have the `reshape` method which exists in `NSOpenGLView`.

falkTX:
@jpcima can you recheck this with develop branch? recently did some fixing for this, on my own tests it all works fine within reaper. even high-dpi is working nicely now.
LV2 is not working correctly, but that is a different matter. (reaper expects LV2 UI resize extensions but I am really trying to go without them for the next DPF)

falkTX:
Closing this issue since I am pretty sure it is resolved now.
Me and another user were able to reproduce the issue, and confirmed it fixed too.

So I assume fixed unless told otherwise.

PR #200, opened 2019-11-14
Fix user-resizable embedded window in Reaper Mac

jpcima:
This will inhibit applying the auto-resize mask until
the initial size has been set in the PuglView. Initially,
DPF would set this size to (1, 1).

It's a very simple solution for #198, too simple maybe.
It needs to verify if it hasn't broken something else.


falkTX:
Looks fine, being a sorta hack and all.
Thanks

PR #201, opened 2019-11-17
Add Shiro-plugins

trebmuh:


falkTX:
This is nice, but it is all quite unsorted...
If we keep adding more, would be nice if names are sorted alphabetically

trebmuh:
Would you mind merging this one, and I'll sort them afterwards in another PR ?

falkTX:
That is quite fair, will merge then

trebmuh:
ty

PR #203, opened 2019-11-25
lv2: require the idleInterface feature

jpcima:
lv2lint 0.2.0 complains about the `idleInterface` feature when it's not requested.
This problem has been reported in Ninjas2: https://github.com/rghvdberg/ninjas2/issues/76

```
    [FAIL]  Idle Interface
              lv2:feature ui:idleInterface missing
              seeAlso: <http://lv2plug.in/ns/extensions/ui#idleInterface>
```


falkTX:
Makes sense.
It needs more than that though...

ui resize, parent and touch are optional features we should be defining in the ttl too

PR #204, opened 2019-11-25
lv2: request the optional ui:parent feature

jpcima:
This adds the `parent` optional feature to ttl, as requested.

>ui resize, parent and touch are optional features we should be defining in the ttl too

The features `resize` and `touch` are already added.


falkTX:
Thanks a lot!

Though my OCD triggers seeing this addition not respecting alphabetical order :)
Can you change that please? so that parent comes before resize?

jpcima:
No problem, I changed it to alphabetic order.

falkTX:
Haha, thanks again. applied :)

PR #208, opened 2019-12-12
CMake support

CapRat:
Hey i added Cmake support. 
There is no need to make a change on the current CodeBase (I just changed the font file in InfoExampleUI.cpp, you can easily revert it.) 
I Added just CMakeFiles to the sourcepath.
At the Moment the CMake Feature is not final. VST, Jack, LV2, LADSPA and DSSI without gui works fine under Linux. 
In Future releases i would like to add Windows(Visual Studio, etc) support and a more stable CodeBase for CMake, so that you can build it with DPF in source and also install it on your pc and include it with find_package(until now there is just support for insource builds).   
If you are fine with adding CMake along to your current build system, i would like also to add Documentation to build the source with cmake in the future. 
I would like to hear what you think of my plans,
greets and regards


falkTX:
That is very welcome yeah, thank you!

But a build system on a project like DPF is very important to get right, so I will have to be picky on this.
(Projects will most likely reuse/copy your cmake rules, if changes are later needed it becomes very difficult for projects to catch up. I had to manually fix a few DPF-dependent projects myself in the past)

If we are able to keep package checks over pkg-config, that would be nice.
finding cairo and jack for example should be just a matter of asking pkg-config if they are available, and where.

Also, something I noticed.. DPF can build for any posix-compliant OS, so the fallback is to X11, not Haiku. Your files assume otherwise

CapRat:
Ah ok, thank you very much for your Feedback.
You are absolutly right. Than lets skip the CMake integration for now, until i fixed the current Problems.
Maybe we could also add Testing. But important Things first. I will fix the given Problems and try to make it stable. 
Thanks for your fast Feedback! ( i will close this pullrequest and open a new one if im ready) 
Best Regards!

CapRat:
Hey i just want to write an Update with the Progress of CMake and hope to get some feedback again. You can still look at the Code at my fork. 
I fixed all your Issues you Mentioned first and added also some hopefuly CMake Features like detecting which Plugin can be build. 
The Base is more solid right now, but still have DSSI issues(with both the make and my new CMake Variant) on Linux, so i think my System have problems with DSSI. 
I Did some FIxex to get MSVC working, which builds nice working vsts. LV2 ist due to its generation process limitted(Encoding Bugs). 
MinGW doesnt seems to work that good. Its build, but istn recognized buy Tracktion HOST.

PR #210, opened 2019-12-21
CMake Support

CapRat:
Hey, im happy to come here again. So i finished the points you ask me todo.
Right now i have run and tested it on Windows (MinGW, Visual Studio 2019)  and Linux(I cant test on a mac, but it could work directly or dont need much changes)
I Also added a littlebit of Documentation. Its rly not an Art but i hope it helps on Understanding.
Im happy to get some feedpack


CapRat:
Sorry, for make you waiting so long, Got some Problems on Mac Compilen, because the use of 2 auto Types use and remove of no except.
I just tried to fix everything.

jpcima:
Hello, do you mind to split this in smaller PRs for different fixes?

This is great to correct some code to operate with the MS compiler, but it is better not mixed with a PR which is strictly on the topic of CMake.
It makes the PR too large and moreover, the fixes look often like trivialities which can be merged without a problem; it's unlike CMake which will likely need a lot of discussion.

Also, you forgot in this new PR to target the `develop` branch instead of the `master` one.

CapRat:
ah ok, i try to do it. Should i close thins PR then?

PR #213, opened 2020-01-10
Bugfixes and Visualstudio compat

CapRat:
Added some bugfixes, independent for CMake refractoring. ( But are needed for compatibility with Visual Studio)


CapRat:
oh no i committed and created a pull request, afterwards i commited in my fork the cmake files, but github tracked it. Im Sorry aggain :/ not ment to do that. but it still all works.
Can we do it just that way? or Should i reopen it again. I mean you can revert cmake stuff easily

falkTX:
You need to put them on separate branches.
In any case, I cannot merge your changes as-is, there is a lot of stuff to work on, for example the allocations during realtime audio functions (which is not allowed, that is why variable arrays were used).
I can cherry-pick a few good commits if that is okay for you, or do you rather me commenting on specific things that need work and merge only when everything is ready?

CapRat:
no shure, cherrypick everything u want. Im sorry for the bad git commits. I didnt work that much with git :)

CapRat:
Also want to add, that the Codechanges not depend on Cmake support, they are just 4 compatibility wiht Visual Studio!

falkTX:
This PR has been superseeded by #278

PR #215, opened 2020-01-27
Send any MIDI message from UI

jpcima:
Followup PR on JACK & VST `sendNote()` implementation

This adds `sendMidi(data, size)`, for arbitrary small messages. (top commit)
I have changed the signature of the note callback in all wrappers, and renamed.

Request of @rghvdberg for SFZ


falkTX:
@rghvdberg what is the usecase for such an API?

rghvdberg:
> @rghvdberg what is the usecase for such an API?

tbh : I totally forgot :-(

Issue #216, opened 2020-02-10
Widget::setVisible(false) doesn't hide widget


ryukau:
Hi.

As title says, calling `Widget::setVisible(false)` doesn't hide the widget in next draw call. Is this intended behavior?

I'm using `NanoWidget` with `DISTRHO_UI_USE_NANOVG` set to 1.

When I changed `dgl/src/Window.cpp` as following, it starts hiding the widget according to `isVisible()`.

```cpp
void onPuglDisplay() // Line 919.
{
    fSelf->onDisplayBefore();

    FOR_EACH_WIDGET(it)
    {
        Widget* const widget(*it);
        if (widget->isVisible()) // Added visibility check.
            widget->pData->display(fWidth, fHeight, fAutoScaling, false);
    }

    fSelf->onDisplayAfter();
}
```

ryukau:
Sorry, above code was wrong. It didn't hide widget.

Changing `dgl/src/NanoVG.cpp` as following hides `NanoWidget` according to `isVisible()`.

```cpp
void NanoWidget::onDisplay() // Line 976.
{
    NanoVG::beginFrame(getWidth(), getHeight());
    onNanoDisplay();

    for (std::vector<NanoWidget*>::iterator it = nData->subWidgets.begin(); it != nData->subWidgets.end(); ++it)
    {
        NanoWidget* const widget(*it);
        if (widget->isVisible()) // Added visibility check.
            widget->onNanoDisplay();
    }

    NanoVG::endFrame();
}
```

falkTX:
nanovg subwidgets in DPF have some small issues that I still have to fix.
This is one of them, so thanks :)

I need to take care of some JACK related things, but afterwards I will take some time to dedicated to DPF

falkTX:
This has been fixed in the pugl-update-v2 branch, soon to be merged into develop.

PR #218, opened 2020-03-24
Add CVPort example plugin

ryukau:
This is my first time to make a pull request on GitHub. So if I'm doing something wrong, please let me know.

This PR adds CVPort example plugin to show a use case of `Plugin::initAudioPort()`. I first thought there's no method to change LV2 port in DPF. I hope this example may helps others to notice this functionality.

Also contains a fix to issue #216.


falkTX:
Thanks for the PR!
Some of the code comments have copy&paste issues (they refer to the parameters example still, which they got copied from)

In any case, I do not like the idea of this new plugin, for it does nothing but bypass CV.
Can we think of something a simple plugin is able to do with audio and CV?
Plugins can mix together audio and cv ports, I think this will be much better to show.
Maybe an audio to cv plugin? something that has 1 audio in, 1 cv out, and the cv out is RMS peak data. Besides the RMS algorithm, the plugin code would be quite simple.

ryukau:
Hi falkTX. Thanks for your feedback.

I'm not familiar with RMS algorithm, so I added sample & hold functionality instead. The plugin now also has 2 inputs to demonstrate mixed usage of audio port and CV port.

ryukau:
Hi. I added new commit to solve issue #235.

- `barBeat` is added to `TimePosition::BarBeatTick`.
- Metronome example is added to show the usage of `barBeat`.

Currently, it only works on LV2 and VST 2. I couldn't figure out how to implement barBeat on JACK, so it's always set to 0.

falkTX:
hmm I do not quite agree with the changes.
The only extra information you need is the "rounding-error" introduced by converting `ticks` to an integer.
If we make `ticks` a double, we can have it with higher precision and thus sorta serving the same as `barBeats`.

ryukau:
Thanks for the feedback.

Yes. If one of the `tick`, `beat` or `bar` becomes double, it's sufficient for my use case. A problem is that `ticksPerBeat` is [not set in VST 2](https://github.com/DISTRHO/DPF/blob/c78671ad5fe5e3fb548bf7bd3f55729e04974bdd/distrho/src/DistrhoPluginVST.cpp#L993). So even if `tick` becomes double, it only works on LV2 for now. If DPF aren't planning to end the support to VST 2, I think changing `beat` to double is a bit better.

Or, what I really would like to know is the frames to next beat from the start of current cycle. So another idea is to add something like `uint32_t framesToNextBeat;`, and use the best precision calculation for each plugin/standalone format.

What do you think?

falkTX:
> Yes. If one of the `tick`, `beat` or `bar` becomes double, it's sufficient for my use case. A problem is that `ticksPerBeat` is [not set in VST 2](https://github.com/DISTRHO/DPF/blob/c78671ad5fe5e3fb548bf7bd3f55729e04974bdd/distrho/src/DistrhoPluginVST.cpp#L993). So even if `tick` becomes double, it only works on LV2 for now.

It will still work regardless. `ticksPerBeat` on LV2 is not a real thing either, but a custom kxstudio property I made that is only available in a few hosts and plugins.



> Or, what I really would like to know is the frames to next beat from the start of current cycle. So another idea is to add something like `uint32_t framesToNextBeat;`, and use the best precision calculation for each plugin/standalone format.

that would still lose precision, because beats do not 100% fit into integers/frames.
and it is best not to make predictions about future events.

slightly related..
sometime later, I want to add sample-accurate time changes to DPF (supported by at least LV2)
This would just be a function plugins can call to request the "next" time, if there is one.
So plugins that only care about the frame-0 time keep working as usual, but smarter plugins can request extra info and be more in sync. This is needed in order to support loop points, where the current buffer overflows the loop and goes back to the beginning.

ryukau:
> It will still work regardless. ticksPerBeat on LV2 is not a real thing either, but a custom kxstudio property I made that is only available in a few hosts and plugins.

Now I see what's going there. tick is derived from barBeat on [LV2](https://github.com/DISTRHO/DPF/blob/c78671ad5fe5e3fb548bf7bd3f55729e04974bdd/distrho/src/DistrhoPluginLV2.cpp#L443) and [VST 2](https://github.com/DISTRHO/DPF/blob/c78671ad5fe5e3fb548bf7bd3f55729e04974bdd/distrho/src/DistrhoPluginVST.cpp#L1006).

The equation used in DPF is:

```c++
double rest = std::fmod(barBeats, 1.0); // Fraction of a beat.
bbt.tick    = rest * bbt.ticksPerBeat;
```

To retrieve barBeat on plugin side:

```c++
double rest    = bbt.tick / bbt.ticksPerBeat;
double barBeat = rest + bbt.beat - 1; // bbt.beat starts from 1.
```

So it works regardless of the value of ticksPerBeat.

Thanks for clarification. I fixed the patch.

falkTX:
That looks good, thanks a lot!

I wonder about the noise/sound generated by the metronome.. does not sound very useful (I did not check/run the code, but I guess it wont sound "proper")
Is it too much work to add some basic sine-wave there instead?

ryukau:
Changed example to output sine waves.

It was generating impulse, so kind of useless. I tend to make example as simple as possible for users to get answer quickly. There's a balance between too simple and too complex, so feedback is welcome.

falkTX:
I imported the changes to TimePosition into develop, so `tick` is now a double.
WIll handle the other changes (added examples) later on.

falkTX:
CVPort and Metronome has been added as examples in DPF develop branch.
Thanks a lot for those!

PR #220, opened 2020-04-06
Fixed bug with dynamic allocation in the String class

soerenbnoergaard:
Hi,

I have been creating [gunshot](https://github.com/soerenbnoergaard/gunshot) - a convolution plugin which stores the impulse response as plugin state in the host so the impulse response files do not get lost in the file system.

On Linux (tested with Bitwig, Mixbus5, Reaper, Waveform 11), I have seen no problem storing the complete impulse response as a base64-encoded string. However, when I cross compile for Windows (tested with Mixbus5, Reaper, Waveform 11), as describe for the [dragonfly](https://github.com/michaelwillis/dragonfly-reverb) plugin, the plugin crashes when saving state (deactivating the plugin, saving the project, etc).

I have isolated the problem to `getState()` which led me to `DistrhoPluginVST.cpp` line 705, i.e. the `effGetChunk` case where key and value pairs are combined to a `String`. The plugin crashes then trying to append my very long base64-value to `tmpStr`.

I traced this further to the operator methods in the `String` class. It seems that the dynamically sized char-arrays do not allocate correctly using the `mingw32`-cross-compiler.

I have therefore changed all dynamically sized char-arrays to a "new-delete" pattern and now I do not get any of the Windows DAWs i mentioned above. I also tried "malloc-free" which worked as well.

### DIFF
--- a/distrho/extra/String.hpp
+++ b/distrho/extra/String.hpp
@@ -598,7 +598,10 @@ public:
         uint i=0, j=0;
         uint charArray3[3], charArray4[4];
 
-        char strBuf[kTmpBufSize+1];
+        char* strBuf = (char*)std::malloc(kTmpBufSize+1);
+        if (strBuf == nullptr)
+            return String();
+
         strBuf[kTmpBufSize] = '\0';
         std::size_t strBufIndex = 0;
 
@@ -652,6 +655,7 @@ public:
             ret += strBuf;
         }
 
+        std::free(strBuf);
         return ret;
     }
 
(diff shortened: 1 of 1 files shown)


falkTX:
Thanks for the investigation.
Makes sense to limit the stack usage here, and allocate memory if needed.
You need to pair new[] calls with delete[] by the way, but in DPF malloc/free is preferred for simple types as it can return null instead of throwing (so we can keep the "noexcept" rule).

For handling files, please note that I am working on something that has host integration (see develop branch of DPF).
The API is not finalized yet, and only works for LV2 at the moment, but it is something that is going to stay.
I want DPF to handle file handling natively, as state.

soerenbnoergaard:
Hi Filipe,

Thank you for your comments. I have updated my changes to use `std::malloc()` and `std::free()` instead of `new` and `delete` and added error handling in case the allocation fails. I have tried to match the style of the existing code, but please let me know if you have any suggestions.

Best regards,
Søren

soerenbnoergaard:
By the way, being able so save files as state sounds awesome! Would be very useful in my case and may others I can imagine.

Best regards,
Søren

falkTX:
This is fixed in the develop branch, with an alternative approach.
Code has been previously tested on the Carla application (DPF String class is a mostly directly copy of CarlaString class)

Issue #228, opened 2020-05-11
Update nanovg libs


jpcima:
Hello, there is some trouble with current libs and it would nice to have an update.

- stb_truetype cannot load fontawesome
- fontstash must be updated along, since the new stb_truetype seems to break it

also it would be extra nice the nanovg update to go along. It introduces some new graphics API.
(one being `nvgRoundedRectVarying`, interesting addition for UI elements)

falkTX:
I thought nanovg was in sorta limbo/maintenance mode?
Did someone else take over?

pdesaulniers:
AFAIK, it's still in "not actively maintained" mode.

With that said, DPF uses a 5 years-old version of nanovg, so we're missing a couple of features / fixes.

jpcima:
Also, and about the main point of the request.
The side libs of nanovg, [stb](https://github.com/nothings/stb) ones and [fontstash](https://github.com/memononen/fontstash) have their own upstream with more updated versions.

By updating these, and not necessarily nanovg, it's the only way to make it accept the OTF fonts.

PR #229, opened 2020-05-11
Update NanoVG

jpcima:
#228
This would update nanovg and side libs from upstream.

Updating Nanovg appeared, though, to introduce a very significant change in my rendering.

**Before** (update fontstash and stb_truetype only)
Before

**After** (update all of nanovg)
After


falkTX:
I don't really see a difference in the rendering.
What am I missing?

It is impossible to review such big PRs. All fine to merge into develop if it just works as usual

jpcima:
The above, on newer NanoVG, draws the text larger by some amount like 10%.

jpcima:
The nanovg update would change the unit of font sizes.
The last edit resets it like before, which is pixel-based as opposed to em-based, so the update should be backward compatible.

PR #236, opened 2020-07-18
Link all plugins to math lib because math functions are used in DGL

zamaudio:
This should resolve any backward compatibility issues with glibc by versioning all math symbols.


zamaudio:
See https://sourceware.org/pipermail/libc-alpha/2020-May/113840.html

falkTX:
Right. This won't be an issue in the kxstudio repo builds, but might still happen for other people.
Easy to add in.

PR #245, opened 2020-12-22
Fix compiler warnings (unused parameters)

JohannesLorenz:


falkTX:
Hi! Thanks for the PR.

Can you please target develop instead of master?
Also, the style used in DPF for unused variables is to put them at the bottom, as otherwise this "useless" code gets in the way of the stuff that matters.
Put it with a comment on top "// maybe unused". There should be many other instances of such usage within DPF.

Thanks again!

JohannesLorenz:
No problem. I opened #246 now for this. Closing this PR.

Issue #242, opened 2020-11-05
branch develop : compile error ‘fMidiEventURID’ was not declared in this scope


rghvdberg:
```
In file included from ../../dpf/distrho/DistrhoUIMain.cpp:26:
../../dpf/distrho/src/DistrhoUILV2.cpp: In member function ‘void DISTRHO::UiLv2::sendNote(uint8_t, uint8_t, uint8_t)’:
../../dpf/distrho/src/DistrhoUILV2.cpp:312:35: error: ‘fMidiEventURID’ was not declared in this scope
  312 |         atomMidiEvent.atom.type = fMidiEventURID;
```

JohannesLorenz:
I did some research on it. The variable got removed in 9d09b5e15b314eba3b96fdd70171e3f1ad1f0072 in ./distrho/src/DistrhoUILV2.cpp .

riban-bw:
I have submitted https://github.com/DISTRHO/DPF/pull/261 which _should_ resolve this.

falkTX:
Bug is fixed on develop branch.

I will make that the default branch soon, in preparation for the pugl rework.

PR #246, opened 2020-12-22
Fix compiler warnings (unused parameters)

JohannesLorenz:


JohannesLorenz:
Unfortunately, I could not test compilation with this due to #242 .

falkTX:
develop has changed quite a lot and this should no longer be an issue.
please confirm when you can, thanks.

falkTX:
This part of the code has now changed, the function is only exposed if it is really going to be used.
Thus this small fix is not needed anymore.
Thanks anyway.

PR #248, opened 2020-12-31
Make knob mousewheel delta independent of window size, fixes #95

### LINKED ISSUE #95: Scalable UIs - mousewheel/drag delta depends on window size
I believe the knobs are calibrated based on the window size.  So when you stretch the window, the amount of delta on the drags/wheels is scaled down.  This needs to be independent of the current window size so that the amount of adjustment stays constant at different window sizes.

zamaudio:


falkTX:
what does that function do?

it fails on macos-old
```
Compiling src/ImageWidgets.cpp (OpenGL variant)
src/ImageWidgets.cpp: In member function 'virtual bool DGL::ImageKnob::onScroll(const DGL::Widget::ScrollEvent&)':
src/ImageWidgets.cpp:615: error: 'copysignf' is not a member of 'std'
```

zamaudio:
Hmm I thought it was from C++98, my bad.  z = copysignf(x, y) takes the magnitude of x with the sign of y and assigns it to z.  In my case i just want to extract +1.f or -1.f from ev.delta.getY() instead of its value.

falkTX (approved):

PR #249, opened 2021-01-17
Allow building projects outside the examples dir

lucianoiam:
This patch allows setting a `DPF_CUSTOM_PATH` in a DPF project Makefile to allow building projects located outside the Examples directory.

/random/path/Makefile:

```
# Makefile for DISTRHO Plugins #
# ---------------------------- #
# Created by falkTX

DPF_CUSTOM_PATH = /usr/local/src/DPF
DPF_CUSTOM_TARGET_DIR = bin
DPF_CUSTOM_BUILD_DIR = build

...
```

I did it out of necessity for a new project I am working on https://github.com/lucianoiam/dpf-webview

Not sure this is the best way to setup a DPF project but it gave me good results so far.

PR #250, opened 2021-01-18
Allow building projects outside the examples dir

lucianoiam:
Supersedes https://github.com/DISTRHO/DPF/pull/249 to target branch develop

Project Makefiles can now define a DPF_CUSTOM_PATH

When setting a DPF_CUSTOM_PATH also DPF_CUSTOM_TARGET_DIR (for binaries) and
DPF_CUSTOM_BUILD_DIR (objects) need to be set.

If the project depends on DPF's graphic library, it has to be built first by
cd'ing to DPF and running make dgl

### DIFF
--- a/Makefile.plugins.mk
+++ b/Makefile.plugins.mk
@@ -7,10 +7,14 @@
 # NOTE: NAME, FILES_DSP and FILES_UI must have been defined before including this file!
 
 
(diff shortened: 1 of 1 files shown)


falkTX:
Thanks!

PR #255, opened 2021-03-23
Extra mouse input for easier knob editing

martinbangens:
I added some extra mouse input for knobs made with ImageWidgets
the changes are:
Middle mouse click will set the default value.
Right mouse click and hold will work like left click + ctrl.
Right mouse click and hold + ctrl will take minimal input (fine tuning)

Should Middle mouse click set default value be removed? Is this a good idea?

### DIFF
--- a/dgl/ImageWidgets.hpp
+++ b/dgl/ImageWidgets.hpp
@@ -150,6 +150,7 @@ private:
 
     int  fRotationAngle;
     bool fDragging;
+    bool fRightClicked;
     int  fLastX;
     int  fLastY;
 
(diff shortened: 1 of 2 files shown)


martinbangens:
We could redesign those both methods so it will calculate the d value within ImageKnob::onMouse (as much as it can) instead of within ImageKnob::onMotion so it don't have to repeat the calculation. It could just check for the Ctrl key within ImageKnob::onMotion and choose a d value that was previously calculated. I dont know how correct this is but I could look into it more tomorrow and make some more commits for this :) try to make it faster

martinbangens:
This gets closed because the changes should go to develop branch

PR #259, opened 2021-04-03
Fixes #258

### LINKED ISSUE #258: Negative enumeration values are rounded up to wrong value
Negative enumeration values of an integer parameter will be rounded up to wrong integer value when creating LV2 ttl, e.g. an enumeration of {-1, 0, +1} will be written to LV2 ttl as {0, 0, 1}.

This is because line 567 of DistrhoPluginLV2export.cpp:
```
const int roundedValue = (int)(enumValue.value + 0.5f);
```
A solution may be to add a check here, e.g.
```
if(enumValue.value < 0.0)
    pluginString += "            rdf:value " + String((int)(enumValue.value - 0.5f)) + " ;\n";
else
    pluginString += "            rdf:value " + String((int)(enumValue.value + 0.5f)) + " ;\n";
```

riban-bw:
Fixes #258 - wrong value detected for negative enums of integer parameters

Simple fixe to check which directon to round integers. Up if positive and down if negative. Although this may not be the optimal mechanism it is only used when creating configuration files, not during runtime.

### DIFF
--- a/distrho/src/DistrhoPluginLV2export.cpp
+++ b/distrho/src/DistrhoPluginLV2export.cpp
@@ -565,10 +565,7 @@ void lv2_generate_ttl(const char* const basename)
                             pluginString += "            rdfs:label  \"\"\"" + enumValue.label + "\"\"\" ;\n";
 
                             if (plugin.getParameterHints(i) & kParameterIsInteger) {
-								if(enumValue.value < 0.0)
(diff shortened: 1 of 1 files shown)


falkTX:
Thanks, but there is a mix of tabs vs spaces in there.
it is also better looking to make this a one-liner change with ternary operator.

These changes out go to into the develop branch first, but usage vs developing branches doesnt work too well for PRs... :thinking:

PR #260, opened 2021-04-03
Bug/258 negative enums

### LINKED ISSUE #258: Negative enumeration values are rounded up to wrong value

riban-bw:
Fixes #258 - wrong value detected for negative enums of integer parameters

Simple fix to check which directon to round integers. Up if positive and down if negative. Although this may not be the optimal mechanism it is only used when creating configuration files, not during runtime.

I deleted the branch and PR against master and created this against develop as requested.
I implemented the ternary one-liner as requested but I think this is more difficult to comprehend and requires extra brackets. Anyway - it is as requested so hopefully satisfactory.

### DIFF
--- a/distrho/src/DistrhoPluginLV2export.cpp
+++ b/distrho/src/DistrhoPluginLV2export.cpp
@@ -1,6 +1,6 @@
 /*
  * DISTRHO Plugin Framework (DPF)
- * Copyright (C) 2012-2020 Filipe Coelho <falktx@falktx.com>
+ * Copyright (C) 2012-2019 Filipe Coelho <falktx@falktx.com>
  *
  * Permission to use, copy, modify, and/or distribute this software for any purpose with
  * or without fee is hereby granted, provided that the above copyright notice and this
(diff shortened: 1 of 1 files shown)


falkTX:
not sure what happened here, but the commits/diff you have here has nothing to do with the proposed changes, it only has reverts of DPF commits.
likely some rebasing or branch switching went wrong.

riban-bw:
How strange! I will close this PR and submit a new one with the correct changes against current develop branch.

PR #263, opened 2021-04-06
Make knob mouse input with ImageWidgets more flexible

### LINKED ISSUE #262: Make knob input with ImageWidgets more flexible
Make knob input easier and faster to edit by adding right and (maybe) middle mouse button. Functionality that lets the users click and drag a knob in 3 ways. If there could be less instructions that repeats in method ImageKnob::onMotion that would also be good.

1. normal, left click (if the user press Ctrl after clicking and holding down mouse it result in minimal input like before)
2. minimal, right click
3. tiny, right click + Ctrl

martinbangens:
Hey this PR contain this changes

- Minimal input (like left click + Ctrl ), with right click mouse button
- Tiny input, with right click mouse button + Ctrl

Fixes #262

### DIFF
--- a/dgl/ImageWidgets.hpp
+++ b/dgl/ImageWidgets.hpp
@@ -150,6 +150,7 @@ private:
 
     int  fRotationAngle;
     bool fDragging;
+    bool fRightClicked;
     int  fLastX;
     int  fLastY;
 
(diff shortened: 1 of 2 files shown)


falkTX:
I am going to hold off on merging this PR, as I want to do some core changes on these classes.
(use private data struct instead of exposing internals, cairo compatible, renaming vars, etc)
The code will be unmergeable soon, but please do not close.
Will keep this one in mind when starting the documentation for them.

PR #265, opened 2021-05-08
Eliminate function cast warnings with OpenGL

jpcima:
It silences a cascade of warnings when building nanovg on windows.


falkTX (approved):
Right, those are quite annoying. Seems good to me.

PR #269, opened 2021-05-21
Add CMake

jpcima:
This work is an attempt to support cmake, in the cleanest way possible.
This is expressed in a moderately modern cmake, and broken down into relevant functions.

The user simply invokes `add_subdirectory(dpf)`, and then `dpf_add_plugin` as indicated.
It is documented in the file where the helper is defined.

Examples are supported except for ExternalUI.

Added also a workflow with GH actions, now Linux only.
It helps in the perspective to add MSVC later, because it's fast on that platform.

EDIT checked working on cmake 3.7.1 and up


falkTX:
This is a great initiative, but I see that it targets usage by other repos and not building the core lib itself.
Not sure if possible, but would be great if the static libs were built as their own step instead of needing to be added as a function inside each plugin subdir stuff.

I dont know cmake too much, but general high level idea:
```
define project
subdir(dgl)
define function to build plugin, links against dgl
subdir(examples)
```
with dgl/cmake.rules defining how to build the static libs

is this possible?

examples/CairoUI/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+# CMake file for DISTRHO Plugins #
+# ------------------------------ #
+
+dpf_add_plugin(d_cairoui
+  TARGETS jack ladspa dssi lv2 vst
```
falkTX, line 5: ladspa is not useful for cairo ui example since it has no gui

examples/FileHandling/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+# CMake file for DISTRHO Plugins #
+# ------------------------------ #
+
+dpf_add_plugin(d_files
+  TARGETS jack lv2 vst
+  UI_TYPE opengl
```
falkTX, line 6: UI_TYPE should default to opengl if not specified. is this possible?

examples/Latency/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+# CMake file for DISTRHO Plugins #
+# ------------------------------ #
+
+dpf_add_plugin(d_latency
+  TARGETS ladspa dssi lv2 vst
```
falkTX, line 5: latency has no UI, so dssi is pointless here (the exposed plugin is not a synth, makes no use of dssi stuff)

examples/States/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+# CMake file for DISTRHO Plugins #
+# ------------------------------ #
+
+dpf_add_plugin(d_states
+  TARGETS jack lv2 vst
```
falkTX, line 5: dssi should be here too

examples/CairoUI/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
...
+  UI_TYPE cairo
+  FILES_DSP
+      CairoExamplePlugin.cpp
+  FILES_UI
+      DemoWidgetBanner.cpp
+      DemoWidgetClickable.cpp
+      CairoExampleUI.cpp)
+
+target_include_directories(
```
falkTX, line 14: do we require this line to be in every single cmake file?
is there a reason it cant or shouldnt be done inside the `dpf_add_plugin` function?

falkTX:
I realize some of those plugin targets were from from DPF itself. I will correct them on the new pugl branch.

jpcima:
> This is a great initiative, but I see that it targets usage by other repos and not building the core lib itself.
Not sure if possible, but would be great if the static libs were built as their own step instead of needing to be added as a function inside each plugin subdir stuff.

It's a possibility and I've considered it. The idea would be to add `dgl-opengl` and `dgl-cairo` as library targets, and mark these targets with the "exclude-from-all" option, such that they're only getting built if something links to them.

The current work went for a slightly different strategy instead, it's to add the library and deps lazily as they are required; this permitted to skip library checks that the projects doesn't need.

examples/FileHandling/CMakeLists.txt
jpcima: When `UI_TYPE` is not given, it designates a plugin without UI.
How else do you suggest to describe a non-UI plugin?

falkTX:
> It's a possibility and I've considered it. The idea would be to add `dgl-opengl` and `dgl-cairo` as library targets, and mark these targets with the "exclude-from-all" option, such that they're only getting built if something links to them.

is cmake smart about this?

on meson we can define an "internal" static library in a way that unless it is used by something else, it is just not built.
not sure if something like that exists on cmake.

examples/FileHandling/CMakeLists.txt
falkTX: By having the list of UI files to build empty.
Current DPF makefiles already do this. the UI_TYPE is only needed when it is != opengl.

https://github.com/DISTRHO/DPF/blob/master/Makefile.plugins.mk#L73

examples/States/CMakeLists.txt
falkTX: this cant be done now, something to resolve later on

examples/CairoUI/CMakeLists.txt
jpcima: This function is able to work in arbitrary source hierarchy, as opposed to the fixed `plugins/<name>` directory. I didn't have really a reason to enforce it so I let it be that way.
Otoh, there wasn't a concept like a "plugin home" directory (and consequently not a default path for "plugin info" includes).
Depending on preference, I can revise the `dpf_add_plugin` interface for such concept.

falkTX: you mean the dir where the cmake file resides (and thus where the dpf function is called from) might be different from the plugin code dir?

jpcima: > might be different from the plugin code dir?

Yes that is supposed to be possible, as currently written. (but not checked)

jpcima:
> not sure if something like that exists on cmake.

I'm sure at 90% that such functionality is hackable.
What I think as the more idiomatic is simply to add a -D option.

examples/CairoUI/CMakeLists.txt
falkTX: Ok that makes sense.
Seems like good practice, and we want to recommend those.
Let those lines stay then :)

jpcima:
Done as suggested, + the build is adapted such that, when configured from DPF itself and not a parent project, libs and examples will have their build enabled by default.

falkTX:
Very nice! Thanks a lot so far.

A few more requests;
1. Add example usage as TL;DR on the top of cmake/DPF-plugin.cmake
2. Add copyright notice on the cmakefile and cmake/DPF-plugin.cmake, feel free to use your own name. license I assume is okay to be the same as dpf itself.
3. the hidden visibility stuff is always enabled, but it should only be so when using release builds, otherwise debug stuff will be missing stuff

I cant comment on the cmake stuff, so I just assume that part is fine :)

jpcima:
> I cant comment on the cmake stuff, so I just assume that part is fine :)

There remain some items, off the top of my head
- windows static linking, I think to add a MSVC build later and take care of all windows matters
- the fast-math option, it will not be set in the Release mode, which is possibly better as such, but could be provided as a `FAST_MATH` option.

> the hidden visibility stuff is always enabled, but it should only be so when using release builds, otherwise debug stuff will be missing stuff

This is not usually an obstacle to debugging ability.

falkTX:
We can start with something that already works okay and then add in more stuff.
MSVC shouldnt work with DPF as-is anyway, due to use of VLAs (there is a relevant PR for this).

We can assume windows builds are done with mingw for now, cmake has MINGW and MSVC vars that are usable here.

I don't think I will use the cmake stuff myself, so I will leave it up to you.

falkTX (approved):

falkTX:
Feel free to merge at any point.
Happy with the results, thanks.

PR #271, opened 2021-05-21
Cmake link directories

jpcima:
Reverts `target_link_directories` which is cmake >= 3.13, and attempt fixing by another method.


falkTX (approved):
WFM

PR #273, opened 2021-05-23
CMake updates for pugl

jpcima:
CMake updates for pugl upstream


falkTX:
I am not sure if the latest X11 stuff is correct, but we need to get this merged so things build again.
I typically just put an #error for both cases (with and without a lib) to verify that indeed it builds as expected.

PR #274, opened 2021-05-23
More class namespacing for macOS

jpcima:
Add the missing classes


falkTX:
Right I forgot to check those, thanks

PR #276, opened 2021-05-24
cmake: generate the macOS namespace for pugl

jpcima:
This implements the pugl namespace in the macOS build with cmake.
In addition:
- for now, disable cairo stuff on macOS (and MSVC)
- add the CI workflow for macOS and CMake
- make build artifacts available, to validate that macOS namespacing works (it does)
- add required changes for macOS to build

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,7 +14,7 @@ env:
   BUILD_TYPE: Release
 
 jobs:
-  build_for_linux:
+  cmake_linux:
     runs-on: ubuntu-18.04
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


falkTX:
Thanks for those, will merge.

Some questions / stuff to verify:
- builds need to use `-Wl,--no-undefined` (default and unavailable on macOS, but should be set for all others)
- win32 needs -mstackrealign
- macos needs `-Wno-deprecated-declarations` besides `-DGL_SILENCE_DEPRECATION=1`

jpcima:
> macos needs -Wno-deprecated-declarations besides -DGL_SILENCE_DEPRECATION=1

Can it be silenced by warning suppressions written in code instead?
I wouldn't want no-deprecated-declarations to affect the user code also.

falkTX:
Right, I only applied it when building DGL, things will be different in cmake.

PR #278, opened 2021-05-24
Build with MS compiler

jpcima:
Permits MSVC builds and fixes compile errors.

Note:
- adds workflow and artifacts
- currently downloads the khronos opengl headers
- lv2_ttl_generator would not generate the basename correctly
- explicit instantiations like `DGL::Rectangle<int>` produces a sea of warnings in the VS IDE.
  It dislikes that these template definitions are split in Geometry/OpenGL files.
  Such methods should be `GraphicsContext::drawRectangle` rather that `Rectangle::draw`.

### DIFF
--- a/distrho/DistrhoUtils.hpp
+++ b/distrho/DistrhoUtils.hpp
@@ -94,12 +94,12 @@ static inline
 void d_debug(const char* const fmt, ...) noexcept
 {
     try {
-        ::va_list args;
-        ::va_start(args, fmt);
+        va_list args;
+        va_start(args, fmt);
         std::fprintf(stdout, "\x1b[30;1m");
         std::vfprintf(stdout, fmt, args);
         std::fprintf(stdout, "\x1b[0m\n");
-        ::va_end(args);
+        va_end(args);
     } catch (...) {}
 }
 #endif
@@ -111,11 +111,11 @@ static inline
 void d_stdout(const char* const fmt, ...) noexcept
 {
     try {
-        ::va_list args;
-        ::va_start(args, fmt);
+        va_list args;
+        va_start(args, fmt);
         std::vfprintf(stdout, fmt, args);
         std::fprintf(stdout, "\n");
-        ::va_end(args);
+        va_end(args);
     } catch (...) {}
 }
 
@@ -126,11 +126,11 @@ static inline
 void d_stderr(const char* const fmt, ...) noexcept
 {
     try {
-        ::va_list args;
-        ::va_start(args, fmt);
+        va_list args;
+        va_start(args, fmt);
         std::vfprintf(stderr, fmt, args);
         std::fprintf(stderr, "\n");
-        ::va_end(args);
+        va_end(args);
     } catch (...) {}
 }
 
@@ -141,12 +141,12 @@ static inline
 void d_stderr2(const char* const fmt, ...) noexcept
 {
     try {
-        ::va_list args;
-        ::va_start(args, fmt);
+        va_list args;
+        va_start(args, fmt);
         std::fprintf(stderr, "\x1b[31m");
         std::vfprintf(stderr, fmt, args);
         std::fprintf(stderr, "\x1b[0m\n");
-        ::va_end(args);
+        va_end(args);
     } catch (...) {}
 }
 
--- a/distrho/DistrhoUtils.hpp
+++ b/distrho/DistrhoUtils.hpp
@@ -33,6 +33,11 @@
 # include <stdint.h>
 #endif
 
+#if defined(DISTRHO_OS_WINDOWS) && defined(_MSC_VER)
+#include <basetsd.h>
+typedef SSIZE_T ssize_t;
+#endif
+
 #if defined(DISTRHO_OS_MAC) && ! defined(CARLA_OS_MAC) && ! defined(DISTRHO_PROPER_CPP11_SUPPORT)
 namespace std {
 inline float fmin(float __x, float __y)
--- a/distrho/extra/String.hpp
+++ b/distrho/extra/String.hpp
@@ -22,6 +22,11 @@
 
 #include <algorithm>
 
+#if defined(_MSC_VER)
+#include <basetsd.h>
+typedef SSIZE_T ssize_t;
+#endif
+
 START_NAMESPACE_DISTRHO
 
 // -----------------------------------------------------------------------
@@ -658,7 +663,7 @@ public:
         uint i=0, j=0;
         uint charArray3[3], charArray4[4];
 
-        char strBuf[kTmpBufSize+1];
+        char* strBuf = (char*)malloc(kTmpBufSize + 1);
         strBuf[kTmpBufSize] = '\0';
         std::size_t strBufIndex = 0;
 
@@ -711,6 +716,8 @@ public:
             strBuf[strBufIndex] = '\0';
             ret += strBuf;
         }
+
+        free(strBuf);
 
         return ret;
     }
--- a/distrho/extra/String.hpp
+++ b/distrho/extra/String.hpp
@@ -656,14 +656,18 @@ public:
             "abcdefghijklmnopqrstuvwxyz"
             "0123456789+/";
 
+#ifndef _MSC_VER
         const std::size_t kTmpBufSize = std::min(d_nextPowerOf2(static_cast<uint32_t>(dataSize/3)), 65536U);
+#else
+        constexpr std::size_t kTmpBufSize = 65536U;
+#endif
 
         const uchar* bytesToEncode((const uchar*)data);
 
         uint i=0, j=0;
         uint charArray3[3], charArray4[4];
 
-        char* strBuf = (char*)malloc(kTmpBufSize + 1);
+        char strBuf[kTmpBufSize + 1];
         strBuf[kTmpBufSize] = '\0';
         std::size_t strBufIndex = 0;
 
@@ -716,8 +720,6 @@ public:
             strBuf[strBufIndex] = '\0';
             ret += strBuf;
         }
-
-        free(strBuf);
 
         return ret;
     }
--- a/distrho/extra/String.hpp
+++ b/distrho/extra/String.hpp
@@ -22,11 +22,6 @@
 
 #include <algorithm>
 
-#if defined(_MSC_VER)
-#include <basetsd.h>
-typedef SSIZE_T ssize_t;
-#endif
-
 START_NAMESPACE_DISTRHO
 
 // -----------------------------------------------------------------------
--- a/distrho/src/DistrhoDefines.h
+++ b/distrho/src/DistrhoDefines.h
@@ -55,7 +55,7 @@
 # if HAVE_CPP11_SUPPORT
 #  define DISTRHO_PROPER_CPP11_SUPPORT
 # endif
-#elif __cplusplus >= 201103L || (defined(__GNUC__) && defined(__GXX_EXPERIMENTAL_CXX0X__) && (__GNUC__ * 100 + __GNUC_MINOR__) >= 405) || __has_extension(cxx_noexcept)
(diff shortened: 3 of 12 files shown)


falkTX:
Thanks for of all.

I am not comfortable applying all these changes as-is, need to know why a few things are the way they do first.

Why does MSVC complain about geometry stuff being in separate files?
The `GraphicsContext` is intentionally left as small as possible, this part should not change.

distrho/DistrhoUtils.hpp
```diff
@@ -94,12 +94,12 @@ static inline
 void d_debug(const char* const fmt, ...) noexcept
 {
     try {
-        ::va_list args;
-        ::va_start(args, fmt);
+        va_list args;
```
falkTX, line 97: why is this specifically needed?

distrho/extra/String.hpp
```diff
@@ -22,6 +22,11 @@
 
 #include <algorithm>
 
+#if defined(_MSC_VER)
```
falkTX, line 25: it is likely that this will be needed in a few other places.
As such it should go to DistrhoUtils.hpp

distrho/extra/String.hpp
```diff
@@ -658,7 +663,7 @@ class String
         uint i=0, j=0;
         uint charArray3[3], charArray4[4];
 
-        char strBuf[kTmpBufSize+1];
+        char* strBuf = (char*)malloc(kTmpBufSize + 1);
```
falkTX, line 666: does MSVC complain here too? a few lines above we can clearly see we limit `kTmpBufSize` to `65536`.
I sugest to make kTmpBufSize always 65536 for MSVC

distrho/src/DistrhoDefines.h
```diff
@@ -74,7 +74,7 @@
 #if defined(__GNUC__)
 # define DISTRHO_DEPRECATED __attribute__((deprecated))
 #elif defined(_MSC_VER)
-# define DISTRHO_DEPRECATED __declspec(deprecated)
+# define DISTRHO_DEPRECATED [[deprecated]] // Note: __declspec(deprecated) it not applicable to enum members
```
falkTX, line 77: comments placed on macros that are to be used with other stuff in front should use /* */ style of comments rather than //

distrho/src/DistrhoUI.cpp
```diff
@@ -40,7 +40,7 @@ Window*     d_lastUiWindow     = nullptr;
 // -----------------------------------------------------------------------------------------------------------
 
 #if DISTRHO_PLUGIN_HAS_EXTERNAL_UI
-UI* createUiWrapper(void* const dspPtr, const uintptr_t winId, const double scaleFactor, const char* const bundlePath)
+UI* createUiWrapper(void* dspPtr, uintptr_t winId, double scaleFactor, const char* bundlePath)
```
falkTX, line 43: const attributes were removed, likely a mistake

distrho/src/DistrhoUI.cpp
```diff
@@ -54,7 +54,7 @@ UI* createUiWrapper(void* const dspPtr, const uintptr_t winId, const double scal
     return ret;
 }
 #else
-UI* createUiWrapper(void* const dspPtr, Window* const window)
+UI* createUiWrapper(void* dspPtr, Window* window)
```
falkTX, line 57: const attributes were removed here too

distrho/src/DistrhoUILV2.cpp
```diff
@@ -289,7 +289,7 @@ class UiLv2
 
         // reserve atom space
         const size_t atomSize = sizeof(LV2_Atom) + msgSize;
-        char         atomBuf[atomSize];
+        char*        atomBuf = (char*)alloca(atomSize);
```
falkTX, line 292: why is this alloca and not malloc?
I understand it uses the stack, mostly wondering when one picks alloca vs malloc (previously in the string case you had malloc)

distrho/src/ladspa/ladspa.h
```diff
@@ -30,6 +30,14 @@
 extern "C" {
 #endif
 
+#ifndef LADSPA_PLUGIN_EXPORT
```
falkTX, line 33: modifying the ladspa header is not needed, and thus should be avoided.
we manually set this symbol for export ourselves.

utils/lv2-ttl-generator/lv2_ttl_generator.c
```diff
@@ -62,31 +68,27 @@ int main(int argc, char* argv[])
 
     if (ttlFn != NULL)
     {
-        char basename[strlen(argv[1])+1];
+        char* normalPath = makeNormalPath(path);
```
falkTX, line 71: It is not easy to follow what all this is doing, some code comments would be welcome

distrho/src/DistrhoUI.cpp
jpcima: It is not a mistake: the declarations and the definitions of these fuctions will get different mangling, when the arguments differ in constness.

distrho/DistrhoUtils.hpp
jpcima: It creates a syntax error at the `::` location.
Making a wild guess, `va` stuff might be macros on the windows platform.

distrho/src/DistrhoUI.cpp
falkTX: but they dont. the const is related to the variable access, not its type.
It is perfectly valid to do something like this:
```
// header stuff
int somefunc(int x);
// implementation
int somefunc(const int x) { return x; }
```

falkTX: I purposefully do not use `const` on the headers to make it simpler/nicer to look at.
But implementation should use const whenever possible.

distrho/src/ladspa/ladspa.h
jpcima: The definition enters in conflict with the declaration, because one of these has dllexport and the other hasn't.

falkTX: huh that is new to me. with gcc we can redeclare functions to change their visibility status.
so we need the same for DSSI as well, right?

distrho/DistrhoUtils.hpp
falkTX: bummer but ok, lets go with removing :: then

distrho/src/DistrhoDefines.h
jpcima: ok, but note that `override` and `final` definitions are formatted the same

falkTX: right, that is my bad.
I will leave them for now as to not generate merge conflicts.

distrho/src/DistrhoUI.cpp
jpcima: I assure you 100%, that on MSVC compiler, contrary to GCC, the constness of parameters affects the mangling.
This mismatches definition and declaration, and this fails the link.

falkTX: I trust you. Just very odd then, because the same "issue" is used through-out the code in DPF.
It is complaining about these in particular for some reason..
Please change the declaration on DistrhoUIInternal.hpp to include the const attribute, instead of removing them here.
Thanks!

distrho/src/ladspa/ladspa.h
jpcima: Yes it applies to DSSI, so I did the relevant edit (if DSSI ever matters on windows platforms)

distrho/src/DistrhoUILV2.cpp
```diff
@@ -289,7 +289,7 @@ class UiLv2
 
         // reserve atom space
         const size_t atomSize = sizeof(LV2_Atom) + msgSize;
-        char         atomBuf[atomSize];
+        char*        atomBuf = (char*)malloc(atomSize);
```
falkTX, line 292: final thing here, atomBuf should be marked const.

jpcima:
> Why does MSVC complain about geometry stuff being in separate files?
The GraphicsContext is intentionally left as small as possible, this part should not change.

I don't think of it as a blocking issue for the moment.
To elaborate, when compiling `Geometry.cpp`, an explicit instanciation is requested for `Rect<int>` and a number others.

In this source, the definition of `Rect<int>::draw` is not available; the compiler warns about not being able to produce the complete instantiation.
Conversely, when you build `OpenGL.cpp`, you have `draw` but you don't have all the rest of geometry methods defined.

This warns once per-method per-class instantiation, which is the origin of the sea of warnings.
It is seen in the build log.

falkTX (approved):

falkTX:
> > Why does MSVC complain about geometry stuff being in separate files?
> > The GraphicsContext is intentionally left as small as possible, this part should not change.
> 
> I don't think of it as a blocking issue for the moment.
> To elaborate, when compiling `Geometry.cpp`, an explicit instanciation is requested for `Rect<int>` and a number others.
> 
> In this source, the definition of `Rect<int>::draw` is not available; the compiler warns about not being able to produce the complete instantiation.
> Conversely, when you build `OpenGL.cpp`, you have `draw` but you don't have all the rest of geometry methods defined.
> 
> This warns once per-method per-class instantiation, which is the origin of the sea of warnings.
> It is seen in the build log.



> > Why does MSVC complain about geometry stuff being in separate files?
> > The GraphicsContext is intentionally left as small as possible, this part should not change.
> 
> I don't think of it as a blocking issue for the moment.
> To elaborate, when compiling `Geometry.cpp`, an explicit instanciation is requested for `Rect<int>` and a number others.
> 
> In this source, the definition of `Rect<int>::draw` is not available; the compiler warns about not being able to produce the complete instantiation.
> Conversely, when you build `OpenGL.cpp`, you have `draw` but you don't have all the rest of geometry methods defined.
> 
> This warns once per-method per-class instantiation, which is the origin of the sea of warnings.
> It is seen in the build log.

then I think we just have to hide that warning for some particular files.
it sounds like a useful warning to have, but the way dpf code is structured is intentional. The image base widgets cpp even has to be included directly in both cairo and opengl so the compiler generates the templated definitions.

so, I am for keeping the code as-is and simply silence this particular set of warnings.

jpcima:
This is the origin of VST crashes.
2021-05-24 18_43_26-DPF (Debugging) - Microsoft Visual Studio

falkTX:
is the submodule up to date?
I did the changes on pugl as you mentioned on IRC

jpcima:
You have the declaration wrong: it's like https://github.com/lv2/pugl/pull/45/files
it reverses `WINAPI  `and `*`

jpcima:
Second, there's this fix that you added in `Window::PrivateData`.
It's wrong in 2 ways.

```
    if (self != nullptr)
        self->onReshape(uwidth, uheight);
```

- pData should be checked for null, not self
- pData null check would fail regardless, because pData is not initialized

falkTX:
My bad, synced to upstream pugl just now on develop branch.
Should be safe to merge this now then?

jpcima:
> Should be safe to merge this now then?

Yes go ahead

falkTX:
> * pData should be checked for null, not self
> 
> * pData null check would fail regardless, because pData is not initialized

where do you reproduce this?

jpcima:
at `Window::PrivateData::onPuglConfigure`

actual crash at `Window::onReshape` by dereferencing pData

falkTX:
I got that, but I mean using what?
Because it is running here at least with wine.

jpcima:
All hosts tried, regardless windows/wine

falkTX:
I think if we just skip init() call during PrivateData constructor and instead call it on the Window class side, it should solve this.
Since you can reproduce this much easier than me, can you try please? thanks

jpcima:
The `init` appears in 4 different overloads of `Window::PrivateData` within different init sequences, it's more easily said than done.

falkTX:
Right, will take care of this soon myself then.

PR #279, opened 2021-05-24
msvc: warning suppressions for template instantiations

jpcima:


falkTX:
In DPF convention is to pad macros with 1 space inside each condition.
So this should be:
```
#ifdef _MSC_VER
# pragma warning(disable:4661)
#endif
```

Also please add small text describing why the warning is there. Otherwise readers will not be able to follow what 4661 means without looking up online.
Thanks

falkTX (approved):

PR #280, opened 2021-05-25
cmake: allow to build on mingw using static libs

jpcima:


cmake/DPF-plugin.cmake
```diff
@@ -366,7 +366,12 @@ function(dpf__add_dgl_cairo)
   target_compile_definitions(dgl-cairo-definitions INTERFACE "DGL_CAIRO" "HAVE_CAIRO")
 
   target_include_directories(dgl-cairo PUBLIC ${CAIRO_INCLUDE_DIRS})
-  target_link_libraries(dgl-cairo PRIVATE dgl-cairo-definitions ${CAIRO_LIBRARIES})
+  if(NOT MINGW)
```
falkTX, line 369: can you reverse this condition? if mingw first, much easier to understand and matches the changes below.
thanks

falkTX (approved):
Thanks.

Please merge as rebase for PRs with only a few changes like this, so we avoid a bunch of merge requests commits.

PR #292, opened 2021-06-18
Mute which output in generate-ttl.sh

lucianoiam:
Was printing /usr/local/bin/realpath on Mac. Guess it is not the intention.

### DIFF
--- a/utils/generate-ttl.sh
+++ b/utils/generate-ttl.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # function not available on some systems
-if ! which realpath >/dev/null 2>&1; then
+if ! which realpath &>/dev/null; then
     function realpath() {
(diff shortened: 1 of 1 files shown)


falkTX:
That is correct.

Though doesnt this syntax work better? `&> /dev/null` less logic and all output goes to null

lucianoiam:
True. What's the best approach for fixing this? a new additional commit or overwriting this one ?

falkTX:
As you prefer, I will squash and merge the commits from the PR if you do multiple ones.

Note that most of the time I am not on a macOS system, so please verify that something works before pushing.

lucianoiam:
Done. Made sure everything compiles fine from scratch on Mac.

falkTX:
Thanks!

PR #293, opened 2021-06-25
Fix build when DGL_FILE_BROWSER_DISABLED is defined

pdesaulniers:
On the develop branch, the build would fail with this error when `DGL_FILE_BROWSER_DISABLED` is defined:

```
../../dpf/distrho/src/DistrhoUIPrivateData.hpp:155:19: error: ‘uiStateFileKeyRequest’ was not declared in this scope
```

### DIFF
--- a/distrho/src/DistrhoUIPrivateData.hpp
+++ b/distrho/src/DistrhoUIPrivateData.hpp
@@ -152,7 +152,7 @@ struct UI::PrivateData {
 
     ~PrivateData() noexcept
     {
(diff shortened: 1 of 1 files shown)


falkTX:
ah right, do you mind adding an extra check for `!DISTRHO_PLUGIN_HAS_EXTERNAL_UI` too? seeing the rest of the code this condition is needed too. thanks!

pdesaulniers:
My bad! Should be good now.

falkTX:
Thanks!

PR #295, opened 2021-06-30
Fix typos in DGL source comments

lucianoiam:


falkTX:
Silly copy&paste hah, thanks!

PR #297, opened 2021-07-04
Simplify custom Makefile paths

lucianoiam:
After fiddling a lot with my own Makefiles and learning some more from make, realized PR https://github.com/DISTRHO/DPF/pull/250 can be reworked in a more elegant / less obtrusive way by simply leveraging operator `?=` at the appropriate places.

This almost reverts the Makefile.plugins.mk to its pre #250 version while retaining that PR functionality.

### DIFF
--- a/Makefile.plugins.mk
+++ b/Makefile.plugins.mk
@@ -7,14 +7,12 @@
 # NOTE: NAME, FILES_DSP and FILES_UI must have been defined before including this file!
 
 
-ifeq ($(DPF_CUSTOM_PATH),)
+ifeq ($(DPF_PATH),)
 ifeq (,$(wildcard ../../Makefile.base.mk))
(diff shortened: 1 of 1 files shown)


falkTX:
If these variables are going to be modifiable externally, please prefix them with `DPF_`.
Otherwise someone that by change has `TARGET_DIR` in their env will have a broken build.
Thanks.

lucianoiam:
> If these variables are going to be modifiable externally, please prefix them with `DPF_`.
> Otherwise someone that by chance has `TARGET_DIR` in their env will have a broken build.

That is absolutely right, added a commit fixing that. So these are the externally modifiable vars now:

DPF_PATH
DPF_TARGET_DIR
DPF_BUILD_DIR

falkTX:
Seems good to me, thanks!

Issue #291, opened 2021-06-17
Wrong dimensions reported to REAPER/Mac


lucianoiam:
Looks like a bug and only happens on Mac. Repeating the same for Linux & Win results in REAPER tightly wrapping the plugin UI.

DPF 92f0f97

<img width="1436" alt="Screen Shot 2021-06-17 at 13 59 16" src="https://user-images.githubusercontent.com/930494/122393824-956cf080-cf75-11eb-8c3c-2bcd1ca9c75f.png">

![Screenshot_2021-06-17_14-15-21](https://user-images.githubusercontent.com/930494/122394729-96525200-cf76-11eb-9adb-f44514a3a794.png)

<img width="960" alt="Capture" src="https://user-images.githubusercontent.com/930494/122396007-d82fc800-cf77-11eb-9d87-1a73ce735671.PNG">

falkTX:
is this because of high-dpi use on mac?

lucianoiam:
Likely yes, the Mac has a HiDPI resolution set.

lucianoiam:
The Windows in the screenshot has system scaling set to 150%, btw

falkTX:
Got it. I have it in the plans to investigate macOS and high-dpi in general.
Dont have a retina display, but have a 4k tv that can simulate it.

For standalones, DPF needs to be able to find the scale factor in use too.
Shouldn't be too difficult on macOS and Windows, biggest pain is on Linux trying to detect scale factor in the many DEs available..

lucianoiam:
The dpf-webui project has a `platform::getSystemDisplayScaleFactor()` function that also works standalone, but I'd really like to switch to DPF's own methods when ready ( assuming there's some WIP there )

lucianoiam:
I believe there is a coordinate space mismatch between DGL and what REAPER expects. Likely the former is including the screen scaling factor in its math, and the latter expects unscaled NSView frame values. Sorry I cannot submit a patch this probably requires non-trivial changes. A workaround is unapplying the scale factor here:

DistrhoUI.cpp

```
void UI::onResize(const ResizeEvent& ev)
{
    UIWidget::onResize(ev);

    const uint width = ev.size.getWidth() / theCurrentScaleFactor; 
    const uint height = ev.size.getHeight() / theCurrentScaleFactor;
    uiData->setSizeCallback(width, height);
}
#endif // !DISTRHO_PLUGIN_HAS_EXTERNAL_UI
```

[...]

PR #301, opened 2021-08-05
Workaround for plugin window size on Mac Retina

lucianoiam:
This is a patch for https://github.com/DISTRHO/DPF/issues/291. It is far from perfect, for example it assumes plugin windows are displayed on the main screen. Have been testing a similar solution for a while in my plugins and could not find negative side effects.

### DIFF
--- a/distrho/DistrhoUI.hpp
+++ b/distrho/DistrhoUI.hpp
@@ -54,6 +54,10 @@ END_NAMESPACE_DGL
 
 START_NAMESPACE_DISTRHO
 
+#ifdef DISTRHO_OS_MAC
+float getMacMainScreenBackingScaleFactor();
+#endif
+
 /* ------------------------------------------------------------------------------------------------------------
  * DPF UI */
 
--- a/distrho/DistrhoUI_macOS.mm
+++ b/distrho/DistrhoUI_macOS.mm
@@ -32,4 +32,15 @@
 #define PuglWindowDelegate DISTRHO_MACOS_NAMESPACE_MACRO(DGL_NAMESPACE, PUGL_NAMESPACE, WindowDelegate)
 #define PuglWrapperView    DISTRHO_MACOS_NAMESPACE_MACRO(DGL_NAMESPACE, PUGL_NAMESPACE, WrapperView)
 
+#import <AppKit/Appkit.h>
+
 #import "src/pugl.mm"
+
+START_NAMESPACE_DISTRHO
+
+float getMacMainScreenBackingScaleFactor()
+{
+    return (float)[NSScreen mainScreen].backingScaleFactor;
+}
+
+END_NAMESPACE_DISTRHO
--- a/distrho/src/DistrhoUI.cpp
+++ b/distrho/src/DistrhoUI.cpp
@@ -184,10 +184,6 @@ void UI::uiFileBrowserSelected(const char*)
 /* ------------------------------------------------------------------------------------------------------------
  * UI Resize Handling, internal */
 
-# if DISTRHO_OS_MAC
-extern float getMacMainScreenBackingScaleFactor();
-# endif // DISTRHO_OS_MAC
-
 void UI::onResize(const ResizeEvent& ev)
 {
     UIWidget::onResize(ev);
--- a/distrho/src/DistrhoUI.cpp
+++ b/distrho/src/DistrhoUI.cpp
@@ -184,12 +184,23 @@ void UI::uiFileBrowserSelected(const char*)
 /* ------------------------------------------------------------------------------------------------------------
  * UI Resize Handling, internal */
 
+# if DISTRHO_OS_MAC
+extern float getMacMainScreenBackingScaleFactor();
+# endif // DISTRHO_OS_MAC
+
 void UI::onResize(const ResizeEvent& ev)
 {
     UIWidget::onResize(ev);
 
+# if DISTRHO_OS_MAC
+    float k = getMacMainScreenBackingScaleFactor();
+    const uint width = ev.size.getWidth() / k;
+    const uint height = ev.size.getHeight() / k;
+# else
     const uint width = ev.size.getWidth();
     const uint height = ev.size.getHeight();
+# endif // DISTRHO_OS_MAC
+
     uiData->setSizeCallback(width, height);
 }
 #endif // !DISTRHO_PLUGIN_HAS_EXTERNAL_UI


falkTX:
shouldnt this be handled directly in pugl?
or at least this needs to go in the window private data class, as we can create windows that are not part of a plugin.
(for example, for an about dialog or preferences, that is not embed in the host; or just standalone tools even)

lucianoiam:
I think the culprit is only a mismatch between the coordinate systems of the DPF plugin (ultimately provided by Pugl, right?) and the host. The former already has the scaling factor applied and the latter expects AppKit, non-scaled pixels, ie. what you would find in a NSView rect (this is an assumption but I cannot think why it should be otherwise). So when the host queries the plugin window size it gets oversized dimensions on Retina. If this is correct then there is no generalized issue, and no need to cover the mentioned additional use cases.

One fact that supports this is that the standalone examples look perfect without the patch applied (and the patch also won't affect them).

lucianoiam:
I am also assuming Pugl pixels are not AppKit pixels but expressed in the OpenGL coordinate system instead. But Pugl and OpenGL are really out of my league. Take the above with a grain of salt.

lucianoiam:
Tried a hacked d_states that popups a 200x200 window when one of the blocks is clicked. See how the black 200x200 window size remains constant for all tested scenarios: standalone, standalone+patch, reaper, reaper+patch. I understand this as proof that the bug only affects the main window and only when running as plugin. And the problem is actually in the container view provided by the host, which is oversized because of the explanation above, there is no problem with the DPF windows themselves.

[ExampleUIStates.cpp.zip](https://github.com/DISTRHO/DPF/files/6980951/ExampleUIStates.cpp.zip)

Screen Shot 2021-08-13 at 10 05 14

Screen Shot 2021-08-13 at 10 06 12

Screen Shot 2021-08-13 at 10 00 36

falkTX:
I dont quite get how `backingScaleFactor()` changes between use in one application and not the other, but alright if it works it works.
Putting it on the UI class is still not ideal though.
Please try moving the special division case to `Window::PrivateData` class, on its `onPuglConfigure` method.
The uwidth/uheight can be modified in there if `isEmbed` is true, which then will make this apply to all Windows that are embed into something else (without them having to be plugins).

falkTX:
Oh and then for moving this to window class we can no longer use the plugin UI methods, so `getMacMainScreenBackingScaleFactor` needs the same treatment regarding pugl.hpp/cpp as you did in the previous PR.
From there we can have access to `NSApplication` and `NSView` of pugl, so maybe there is a way to access the current screen of the view, instead of referencing the "mainScreen".

lucianoiam:
> I dont quite get how `backingScaleFactor()` changes between use in one application and not the other, but alright if it works it works.

That is not quite what is happening but I'm sure it is my fault for not properly presenting the issue.

These snippets from Pugl are illustrative:

https://github.com/lv2/pugl/blob/c98ee490fcf93e290ae5e8a602fd88b359a61a1b/src/mac.m#L1322

```C
PuglStatus
puglSetFrame(PuglView* view, const PuglRect frame)
{
  PuglInternals* const impl = view->impl;

  // Update view frame to exactly the requested frame in Pugl coordinates
  view->frame = frame;

  const NSRect framePx = rectToNsRect(frame);
  const NSRect framePt = nsRectToPoints(view, framePx);
  if (impl->window) {
    // Resize window to fit new content rect
    const NSRect screenPt = rectToScreen(viewScreen(view), framePt);
    const NSRect winFrame = [impl->window frameRectForContentRect:screenPt];

...
```

The Pugl coordinates mentioned in the code comment above ( used throughout the Pugl APIs and hence DPF ) are already multiplied by the backing scale factor. See how the `PuglRect` argument values are divided by the scale factor here `  const NSRect framePt = nsRectToPoints(view, framePx)`  before passing them to AppKit later in `puglSetFrame()`.

https://github.com/lv2/pugl/blob/c98ee490fcf93e290ae5e8a602fd88b359a61a1b/src/mac.m#L57

```C
static NSRect
nsRectToPoints(PuglView* view, const NSRect rect)
{
  const double scaleFactor = [viewScreen(view) backingScaleFactor];

  return NSMakeRect(rect.origin.x / scaleFactor,
                    rect.origin.y / scaleFactor,
                    rect.size.width / scaleFactor,
                    rect.size.height / scaleFactor);
}
```

_By the way this scale factor is read from the view's attached screen which is the correct thing to do instead of reading the main screen scale, as you pointed out_

Here is a hacked pugl.mm that serves as proof that on Retina, DPF coordinates != AppKit coordinates:

```C
PuglStatus puglSetWindowSize(PuglView* const view, const uint width, const uint height)
{
    view->defaultWidth  = width;
    view->defaultHeight = height;

#if defined(DISTRHO_OS_HAIKU) || defined(DISTRHO_OS_MAC)
    // keep upstream behaviour
    const PuglRect frame = { view->frame.x, view->frame.y, (double)width, (double)height };
    PuglStatus status = puglSetFrame(view, frame);

    NSWindow *macw = (NSWindow *)puglGetNativeWindow(view);

    printf("Pugl %.f x %.f \n", view->frame.width, view->frame.height);
    printf("OS   %.f x %.f \n", macw.frame.size.width, macw.frame.size.height);

...
```

Running d_states:

```
Pugl 512 x 512 
OS   256 x 256 
```

My understanding of this is that Pugl/DPF thinks in "raw" pixels instead of the OS/AppKit "virtual" pixels hosts expect (tested on REAPER and Live). This sounds like a bold statement but the host essentially provides a NSView for drawing and these use "virtual" pixels, so why not.

First of all already tried moving this to `Window::PrivateData::onPuglConfigure` at no avail. However I insist we should not try to fix this anywhere but in the touching points between DPF and the host. Not sure `DistrhoUI.cpp` is the best place though.

Here is a link to Apple docs with methods that might be cleaner that tinkering manually with the scale factor like initially proposed https://developer.apple.com/documentation/appkit/nswindow/1419603-convertrectfromscreen

falkTX:
> First of all already tried moving this to `Window::PrivateData::onPuglConfigure` at no avail. However I insist we should not try to fix this anywhere but in the touching points between DPF and the host. Not sure `DistrhoUI.cpp` is the best place though.

I am not sure I understand why on this point...
Let's take the example of VST2, where the plugin/UI is in charge of declaring the size to use in the window.
When the host requests `effEditGetRect`, DPF is giving the UI width and height.
The plugin is the one that tells the host what size to use, not the other way around.

The `UI::onResize` is called through the `onResize` virtual function all the way down from the `Widget` class.
This is called during a `setSize()` on the UI/TopLevelWidget.
Without debugging it is not clear what part of the code calls this, but there some calls to it right from the UI constructor.

Seems to me, though I can be wrong of course, that this only works due to DPF calling a `setSizeCallback` that will feed into the host. But this is slightly wrong, the host window should be correct when it requests `effEditGetRect` and not need a set-size right after creating the window. It can be a race condition.

Now, thinking more about this your approach seems correct yes.
BUT the host must still be receiving the wrong initial size, and I think it is important to get that right.
So IMO we need to try to make the host window size correct when commenting out `uiData->setSizeCallback()` (that is, the initial window size), and only then proceed with this patch.

falkTX:
btw, is the behaviour exactly the same for the LV2 version?

I still need to try this...

lucianoiam:
> > First of all already tried moving this to `Window::PrivateData::onPuglConfigure` at no avail. However I insist we should not try to fix this anywhere but in the touching points between DPF and the host. Not sure `DistrhoUI.cpp` is the best place though.
> 
> I am not sure I understand why on this point...
> Let's take the example of VST2, where the plugin/UI is in charge of declaring the size to use in the window.
> When the host requests `effEditGetRect`, DPF is giving the UI width and height.
> The plugin is the one that tells the host what size to use, not the other way around.

Thanks for the clarification. In the case of VST2 that is exactly where the coordinates mismatch occur. Using this argument I would expect the same for any other format but for LV2 results are different, more on below.

> 
> The `UI::onResize` is called through the `onResize` virtual function all the way down from the `Widget` class.
> This is called during a `setSize()` on the UI/TopLevelWidget.
> Without debugging it is not clear what part of the code calls this, but there some calls to it right from the UI constructor.
> 
> Seems to me, though I can be wrong of course, that this only works due to DPF calling a `setSizeCallback` that will feed into the host.

This is actually correct. The issue goes away when passing "native" dimensions to `setSizeCallback()`

 > But this is slightly wrong, the host window should be correct when it requests `effEditGetRect` and not need a set-size right after creating the window. It can be a race condition.
> 
> Now, thinking more about this your approach seems correct yes.
> BUT the host must still be receiving the wrong initial size, and I think it is important to get that right.
> So IMO we need to try to make the host window size correct when commenting out `uiData->setSizeCallback()` (that is, the initial window size), and only then proceed with this patch.

Wouldn't this break resizing the plugin window? note I don't see any direct calls to `setSizeCallback()` in the UI constructor but only in `UI::onResize()`:

>btw, is the behaviour exactly the same for the LV2 version?

>I still need to try this...

With latest official Carla-2.3.2-macOS-universal, LV2 plugin content view tightly fits its window but the window itself looks oversized. The patch has no effect for LV2. I'm puzzled, does this make sense to you?

Oversized is subjective but based on how d_states looks relative to the desktop size on Linux I would say there is an issue:

Screen Shot 2021-08-13 at 19 09 53

Screen Shot 2021-08-13 at 19 09 46

Note VST2 is broken on Carla, plugin keeps resizing itself to exaggerate dimensions back and forth, and closing the plugin window ultimately crashes Carla. But that deserves a separate conversation.

falkTX:
With carla the issue might be 2x double scaling, as carla passes the desktop/ui scaling factor to the plugin UI.
I am trying to setup something here where I can reproduce these issues.
We can focus on getting things to work nicely with standalones first (with and without desktop scaling), and then move to VST2 **without** the setSizeCallback stuff (to ensure the initial size the plugin gives to the host is correct). We might need to tweak things on the VST2 specific code, that will be fine.

falkTX:
Ok let's do this. just added something to fetch desktop scale factor in 9b9e0a36187a5a2f76add523f1eadfe8bf22dbdc
This can be used in standalones and windows created outside of host control.
I adjusted d_info so that it follows scale factor right from the start, this one is better for use compared to d_states since it has text and it will be easier to verify visually if correct or not.

My setup is using a 4k tv screen at 720p scale (so 4x). I confirmed it working correctly for standalones.
this is the result:
image

can you confirm latest commit makes d_info work ok for you in standalone mode?

lucianoiam:
Here running a MacBook Pro Retina, system info says 13,3-inch (2560 × 1600). No external monitor attached.

d_info looks better than before, I recall font was microscopic a couple of commits ago. However the difference compared to Linux is noticeable:

Screen Shot 2021-08-13 at 21 24 40

Screen Shot 2021-08-13 at 21 25 43

I guess the expected result is the one on Linux, around 1/3 of the screen height. On Mac it looks closer to 3/4. This makes me think whether there is an issue for standalones to start. Which is the expected result in your opinion?

falkTX:
There is no implementation to get scale factor on linux, so it will always be 1 for now. (in standalones)
On your screenshot the linux one looks correct, font size is similar to desktop font size.

For reference, this is the expected size on a 1080p screen:

Screenshot_20210813_203944

and just the window: (size 445x296)

Screenshot_20210813_204014

Curiously the size on linux doesnt match the size set by the plugin UI (which is 405x256)

The macOS one does indeed look too big.
I think I was wrongly assuming it was correct. But it is weird, because when manually setting scale to 4x in the code I got the same results as in my screenshot... still ends up looking 2x too big.
Doing some quick calculations 4x plugin UI size is 1620x1024 - I calculate about 1616x1022px on the screenshot.
So even though it looks quite big in my case, it is still correct.

Do you have your desktop set as "alike" 720p or 1080p?
The size of the topbar in your screenshot is the same as mine, so I assume you did 720p to mimic my results?
If that is the case, then it looks to be correct.

falkTX:
ah I see what is wrong with the screenshot of the linux UI, kwin likes to add fancy shadows. they must be around 25px in each side.
so we can assume the linux UI size is correct

falkTX:
an extra note: something is wrong with the way d_info scales up with the factor >1, but that is a separate issue.
something like `double infoScaleFactor = 1.0 + (scaleFactor - 1.0) / 2;` likely will make the size fit better with >1 factors.

I am going to check with carla + lv2 next, since I have control over both sides for that.
Then carla + vst2, and finally other hosts.

When the macOS side works 100%, the rest should be pretty easy.
Except a nice way to find scale factor under x11... ay ay

lucianoiam:
Just realized this Mac display was slightly scaled up, however setting the default does not make a lot of difference. Though that might explain why top bars were the same size...

Screen Shot 2021-08-13 at 22 17 35

I am quite sure there are lots of places where to check, please let me know how can I help here. I think this requires some more knowledge of the internals for all the parts involved, my reasoning has been more intuitive so far. But I am happy to test / write experimental code / etc to get this fixed.

This and https://github.com/lv2/pugl/pull/57 are kind of stoppers on Mac . Just saw there is some very recent progress in that PR !

Besides these everything runs sweet on macOS.

For X11 is it too naive to have a function that gathers as much scaling "sources" as possibly and picks one based on some heuristics? or maybe decide based on a compile time setting. I found GDK_DPI_SCALE to be effective...in my particular setup, lol.

falkTX:
> For X11 is it too naive to have a function that gathers as much scaling "sources" as possibly and picks one based on some heuristics? or maybe decide based on a compile time setting. I found GDK_DPI_SCALE to be effective...in my particular setup, lol.

Yeah, I think this is something you can help a lot.
We can try to fetch info from as many sources as possible. We should even read the Qt test env vars IMO.
The scale factor on this case is only relevant for standalones, but with the work on RtAudio they might become quite nice soon.

On the macOS side, yeah I am following that issue closely myself. It is a tricky one..
Worst case we keep a patch against pugl while drobilla hasnt merged anything officially.
tbh I still consider current/develop DPF to be "not quite ready yet", but it is very close. Getting these final issues out of the way is crucial.

btw, I found out what was wrong with d_info.. I totally forgot I already handle all the needed scaling behind the scenes in DPF when UIs setup geometry constraints.
so the new code was scaling things 2x on top of the desktop 2x scaling.
It now looks correct on my macOS setup:

image

please give it a try, let me know if it looks proper for you too. (standalone)

falkTX:
oh I added a scale factor print on the gui as well, very handy

lucianoiam:
This looks perfect now, and reported scale factor is correct (2.0). Very promising!

Screen Shot 2021-08-13 at 23 08 16

I have been in almost daily contact with plugins produced out from the develop branch for several weeks now and it might not be ready for production but feels very close. Cannot recall any other serious issues, I wrote down two for Windows but those are not stoppers at all (will comment later on that).

For X11 yes I recall Qt relies on its own env vars. I came across a couple of different env vars while trying to get things to look right on a 13" 1080p display ( what a challenge ). I learnt there is no standardized approach for scaling. Will start compiling a list.

falkTX:
Alright, standalone and VST2 should be working well now.
I went with my idea of scaling the size before passing it to the host. Tested to work on REAPER, Renoise and Carla bridges.
Carla itself still has problems, but I think I know why.

Also, REAPER had flickering resize (UI would go on top of the UI buttons sometimes), this is fixed too.

LV2 is a harder one to tackle, REAPER does not support the extension-less approach that I am pushing with DPF.
I want to get all working within Carla first before bothering REAPER devs. (not just resize but min-size and aspect ratio from host side)

falkTX:
The resize issue on main Carla is solved now (Carla-side issue, nothing to do in DPF)

lucianoiam:
This is huge, thanks! also noticed that tiny bug on REAPER that made the UI display on top of the buttons, here it happened when resizing a certain plugin to the max. Minor but glad it is gone.

I tried four examples and all the size issues on Mac are gone. Excepto for d_cairoui, maybe Cairo needs special treatment?

Screen Shot 2021-08-14 at 08 53 00

falkTX:
The cairo side should be fixed now too, it was a good test as it showed a few other bugs, all fixed now.

falkTX:
Closing this ticket since the issues on macOS are resolved now.
For LV2 we can tackle / discuss on another place.

Still need to do some testing for Windows, we need a way to ask for global/screen/desktop scaling for standalones.

lucianoiam:
Just tested; it looks great now. I have some code for Linux and Windows, will improve the former to include some more "sources" like Qt and then open a new issue. Thanks.

PR #302, opened 2021-08-05
Fix Mac standalone window focus

lucianoiam:
CI build is failing but seems unrelated to this patch

```
-- Checking for module 'liblo'
38
--   No package 'liblo' found
39
-- Configuring incomplete, errors occurred!
40
See also "/Users/runner/work/DPF/build/CMakeFiles/CMakeOutput.log".
41
Error: Process completed with exit code 1.
```

### DIFF
--- a/dgl/src/ApplicationPrivateData_macOS.mm
+++ b/dgl/src/ApplicationPrivateData_macOS.mm
@@ -1,0 +1,29 @@
+/*
+ * DISTRHO Plugin Framework (DPF)
+ * Copyright (C) 2012-2021 Filipe Coelho <falktx@falktx.com>
+ *
+ * Permission to use, copy, modify, and/or distribute this software for any purpose with
+ * or without fee is hereby granted, provided that the above copyright notice and this
+ * permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES WITH REGARD
+ * TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS. IN
+ * NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL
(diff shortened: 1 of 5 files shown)


dgl/src/ApplicationPrivateData_macOS.mm
```diff
@@ -0,0 +1,29 @@
...
+#import <AppKit/AppKit.h>
+
+#include "ApplicationPrivateData.hpp"
+
+START_NAMESPACE_DGL
+
+void macActivateApp()
+{
+    [NSApp setActivationPolicy:NSApplicationActivationPolicyRegular];
+    [NSApp activateIgnoringOtherApps:YES]; // focus window
```
falkTX, line 26: this is called on other places in pugl, no? I dont think this line in specific is needed.

falkTX:
Thanks for the PR, but the setup with these files is not good.
Please put the new function in pugl.hpp/cpp instead, there are already a couple of OS-specific functions there.

falkTX:
Actually maybe this needs to be in pugl directly?
Is there a reason not to?

lucianoiam:
Thanks for reviewing. if the standalone plugins were built as app bundles (akin to the VST2 bundle) this call wouldn't be necessary. The ultimate fix is providing a .app bundle. I don't think Pugl should assume how the binaries are to be distributed (bare executable vs. bundle). This is more like a stopgap measure. 

Would it be acceptable to move `macActivateApp()` to pugl.mm, get rid of the new .mm file and submit a new PR ?

falkTX:
> Would it be acceptable to move `macActivateApp()` to pugl.mm, get rid of the new .mm file and submit a new PR ?

Yes, just name it in the same style as the existing extra functions stuff.

I think the issue should be brought to pugl too in any case.
That call doesn't hurt if running through an app bundle, yes?

lucianoiam:
If this is moved to Pugl, both NSApp methods should be called once during application startup. Pugl currently calls `activateIgnoringOtherApps` in `puglRealize()`, if I correctly understand that is each time a new window is created, not sure about the rationale behind that.

Just tested and patch does not hurt when running as bundle, here is a quick test showing that and also how wrapping binary in .app fixes the issue without needing the NSApp calls.

[d_states.zip](https://github.com/DISTRHO/DPF/files/6940493/d_states.zip)

lucianoiam:
Also note Pugl does not call `[NSApp setActivationPolicy:NSApplicationActivationPolicyRegular]` which is also needed.

PR #303, opened 2021-08-05
Rework PR #302 in a cleaner way

lucianoiam:
Replaces https://github.com/DISTRHO/DPF/pull/302

- Patch implementation moved to pugl.mm alongside other platform-specific code
- Renamed function to follow existing convention
- Avoid unnecessary new file and changes in makefiles

This update will become obsolete but safe to keep once DPF provides application bundles for the standalone JACK target.


falkTX:
That looks nice and clean, thanks a lot!

PR #309, opened 2021-08-20
Remove the const qualifier on ParameterEnumerationValue

jpcima:


falkTX:
So this bothers you enough haha.
Feel free to merge after/if the CI tests all pass.

falkTX:
err, they dont. some of the code needs adjusting now.
which makes me think this will break backwards compatibility...

jpcima:
It's how you see fit. Admittedly, I don't need it to have it that much.

falkTX:
It makes the code cleaner, so it is nice to have.
I see the issue was not on the plugin code, this is the part we need to watch out to not break.
Internals of DPF can change of course. So I will merge this, thanks.

PR #310, opened 2021-08-21
Add missing call to FreeLibrary()

lucianoiam:
It was not being called in case of success


falkTX:
right, good catch. please dont open brackets inline though, it is not consistent with the rest of the file.

lucianoiam:
You're right, fixed!

falkTX:
Thanks!

PR #316, opened 2021-09-01
Partial UI_TYPE=external support on Windows

lucianoiam:
This patch allows external UI plugins to build on Windows while external implementation is completed.

To be fair support has always been there, it was disabled likely because external processes are not implemented for Win32 yet. In my case I don't need such, only a raw HWND to stick something on top, everything happens in-process.

Plugin target works (only tested VST2), standalone is missing from the EmbedExternalUI example.

There are some glitches while resizing but the whole resizing things needs to be revised for all platforms, I know external UI is still WIP  :)

### DIFF
--- a/distrho/extra/ExternalWindow.hpp
+++ b/distrho/extra/ExternalWindow.hpp
@@ -19,9 +19,7 @@
 
 #include "String.hpp"
 
-#ifdef DISTRHO_OS_WINDOWS
-# error Unsupported platform!
-#else
+#ifndef DISTRHO_OS_WINDOWS
 # include <cerrno>
 # include <signal.h>
 # include <sys/wait.h>
@@ -106,9 +104,10 @@ public:
     */
     virtual bool isRunning() const
     {
+#ifndef DISTRHO_OS_WINDOWS
         if (ext.inUse)
             return ext.isRunning();
-
+#endif
         return isVisible();
     }
 
@@ -119,7 +118,11 @@ public:
     */
     virtual bool isQuitting() const
     {
+#ifndef DISTRHO_OS_WINDOWS
         return ext.inUse ? ext.isQuitting : pData.isQuitting;
+#else
+        return pData.isQuitting;
+#endif
     }
 
    /**
@@ -259,9 +262,10 @@ public:
     {
         pData.isQuitting = true;
         hide();
-
+#ifndef DISTRHO_OS_WINDOWS
         if (ext.inUse)
             terminateAndWaitForExternalProcess();
+#endif
     }
 
    /**
@@ -357,15 +361,24 @@ protected:
 
     bool startExternalProcess(const char* args[])
     {
+#ifndef DISTRHO_OS_WINDOWS
         ext.inUse = true;
 
         return ext.start(args);
+#else
+        (void)args;
+        return false; // TODO
+#endif
     }
 
     void terminateAndWaitForExternalProcess()
     {
+#ifndef DISTRHO_OS_WINDOWS
         ext.isQuitting = true;
         ext.terminateAndWait();
+#else
+        // TODO
+#endif
     }
 
    /* --------------------------------------------------------------------------------------------------------
@@ -414,6 +427,7 @@ private:
     friend class PluginWindow;
     friend class UI;
 
+#ifndef DISTRHO_OS_WINDOWS
     struct ExternalProcess {
         bool inUse;
         bool isQuitting;
@@ -510,6 +524,7 @@ private:
             }
         }
     } ext;
+#endif
 
     struct PrivateData {
         uintptr_t parentWindowHandle;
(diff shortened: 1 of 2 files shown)


falkTX:
Sounds good, great to know it works on windows too.
Your html-view project is something we perhaps need to see if can be integrated directly. Specially after the external-ui support is more complete.
Making 3 custom platform-specific codepaths is not something I am very fond of, so something like bringing a native web view sounds much better tbh. there are not many other things that work cross-platform in the same way.

lucianoiam:
Continued here https://github.com/DISTRHO/DPF/discussions/290

lucianoiam:
IMHO there are two separate concepts here, external UI where it makes sense to have external processes, and embedded UI which is what the web view needs and probably LDPF, correct me if I am wrong. I call to reflect this in code:

- Rename UI_TYPE=external to UI_TYPE=native (or window, or embed).
- Move ExternalProcess to its own source file
- Rebuild UI_TYPE=external on top of the UI_TYPE in the first point, this type would depend on ExternalProcess.

falkTX:
IMO external is a good name, as it is something external to DPF.
because this can be anything, on some systems it can even be out of process. is that confusing?

osch:
Yes [LDPF](https://github.com/LDPF/LDPF) only needs external embedded UIs. It has a similiar aim than your web view approach: to have a scripting language to make it easier building UIs.

IMHO  the name UI_TYPE="external" is as good or bad as "native", because both could be misinterpreted. "external" could mean "outside of DPF" and it could mean "external process". "native" has a lot of meanings....

lucianoiam:
I agree all choices depend on context, I would defer that to the maintainer. But what about separating both features in code? embed UI support and external process mgmt

falkTX:
right, the external process can be a class on top of the external window one, ExternalWindow -> ExternalProcessWindow or something.

we need to also think about if this class is ever useful on its own, so that we split in 3 actually:
- ExternalWindow
- ExternalProcess
- ExternalProcessWindow

I wonder if external windows are useful as generic classes for DPF UI usage (ie, if you have an OpenGL UI and want an external window?). Not sure if that makes sense or only complicates things further.
Unless there is a clear usecase for such external-window class, I would move it out of dpf/distrho/extra into where DistrhoUtils.hpp resides, perhaps directly call it ExternalUI instead of ExternalWindow...

lucianoiam:
I can think of two use cases:

1. A webview plugin wanting to popup a further webview in a floating window. For example for implementing a config dialog. That would not be a nice UI but it is objectionable.

2. A DGL-based plugin wanting to popup a window to host some system or 3rd-party provided native view, like but not necessarily a webview. A bit exotic though.

Tbh, I also like the idea of just renaming the class to ExternalUI. It would only leave out some rare cases.

lucianoiam:
I think my last comment only makes sense if ExternalWindow was actually able to create system windows, which is not the case, sorry for the confusion.

lucianoiam:
I am having this issue on Windows, see how the plugin content obscures REAPER's view at the top. I am quite sure there was a similar issue affecting DGL (on Mac though, I think) which was corrected. Any idea on this?  is it reproducible in LDPF ?
 
https://user-images.githubusercontent.com/930494/131867251-a9993eec-d335-46a5-b255-6ad304bdbab9.mov

falkTX:
print the value of vst rect "top" variable, perhaps it is not zero.
the macos fix from before was to do reparenting and resizing without forcing the top and left coordinates as zero. I basically swapped "set pos and size" for a "set size" and everything just works.
likely you will need something similar, dont assume window starts at 0,0

lucianoiam:
Thanks, makes sense. Tried that but unfortunately was not the solution, finally it was my ignorance of the Win32 API.

Swapped:

`MoveWindow(fHelperHwnd, 0, 0, width, height, true)`  [ you were right about 0,0 ]

with:

` SetWindowPos(fHelperHwnd, 0, 0, 0, width, height, SWP_NOOWNERZORDER | SWP_NOMOVE)`

And the issue went away. `SWP_NOMOVE` keeps the child window in the same mistery position Windows wants it to be. In other words it only resizes the window.

lucianoiam:
> print the value of vst rect "top" variable, perhaps it is not zero.
> the macos fix from before was to do reparenting and resizing without forcing the top and left coordinates as zero. I basically swapped "set pos and size" for a "set size" and everything just works.
> likely you will need something similar, dont assume window starts at 0,0

On Mac the same was happening and ^^^ was the solution, thanks!

So at this point my non-DGL branch is on par with the original DGL-based version, except for standalones that don't bother right now (I have some draft Mac code for it). It is working really well on the three platforms.

The only remaining "visible" thing is resizing, to recap you mentioned we need to hook into the VST2 callback for that except for REAPER that polls dimensions on its own.

PR #317, opened 2021-09-02
macOS: don't include pugl.mm for external UIs

osch:
On macOS: don't include pugl.mm for external UIs

### DIFF
--- a/Makefile.plugins.mk
+++ b/Makefile.plugins.mk
@@ -89,8 +89,10 @@ endif
 OBJS_DSP = $(FILES_DSP:%=$(BUILD_DIR)/%.o)
 OBJS_UI  = $(FILES_UI:%=$(BUILD_DIR)/%.o)
 
+ifneq ($(UI_TYPE),external)
 ifeq ($(MACOS),true)
 OBJS_UI += $(BUILD_DIR)/DistrhoUI_macOS_$(NAME).mm.o
+endif
 endif
 
 # ---------------------------------------------------------------------------------------------------------------------


distrho/DistrhoUI_macOS.mm
```diff
@@ -14,6 +14,8 @@
  * CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  */
 
+#if !DISTRHO_PLUGIN_HAS_EXTERNAL_UI
```
falkTX, line 17: this macro is at the wrong place, there is no includes ever here, so this for sure will break normal build.

I am not sure if this is the correct approach anyway, we need to see if we can skip this file altogether in the makefiles.

osch: ah yes you are right, I overlooked this (I had the define in the Makefile). So we need a different approach.... The problem is, that the defines are in the header. So perhaps we just could include the header with the defines here before the `#if`?

falkTX: There is a UI_TYPE=external in the makefile for the plugin that is needed, we can just use that as comparison

osch: ah yes, I forgot this... I will then move the `if` to the Makefile....

falkTX:
yeah like that, but it is better to keep the OBJS_UI stuff together, you moved them to a place where it doesnt belong IMO.
the old place seemed better, we can just add another ifneq in there

osch:
I first had the `if` in line 92 but moved it below for two reasons:
1.) UI_TYPE is evaluated and modified before line 175, 
2.) OBJS_UI is also modified below in line 199

But if you still want I can move the `if` back to the top at line 92. IMHO both places have their pros and cons.

falkTX:
Thanks, and sorry being so picky.

Issue #313, opened 2021-08-26
Progress on UI_TYPE=external


lucianoiam:
Great progress there, not sure these are issues or pending features. Thought it is still useful for tracking:

(All macOS for now)

- Running make on EmbedExternalUI does not create the .vst bundle
- With a custom plugin, VST2 works as expected on REAPER, but there is no plugin window at all on Live. Checked that getParentWindowHandle() != 0.
- Any chance of having something like getApp().isStandalone() without DGL?

lucianoiam:
I have been updating all my stuff that runs on top of DPF and `external` is working great so far. On Linux, initial sizing works but then subsequent resizing does not work on Carla and Bitwig (floating windows). On REAPER it works as expected (embedded window). Same binary with configuration:

```
#define DISTRHO_PLUGIN_HAS_UI          1
#define DISTRHO_PLUGIN_HAS_EMBED_UI    1
#define DISTRHO_PLUGIN_HAS_EXTERNAL_UI 1
```

falkTX:
for the resize on vst2, we need to put in place something to report that change to the host.
reaper requests the vst2 plugin window size every second and adjusts to it, but hosts dont typically do this.

lucianoiam:
Makes sense, that perfectly explains what I'm seeing on REAPER when resizing, it works but takes the host a little while to catch up.

lucianoiam:
I tried something like this, inspired on the DGL version:

DistrhoUI.cpp

```
#if DISTRHO_PLUGIN_HAS_EXTERNAL_UI

void UI::sizeChanged(uint width, uint height)
{
    UIWidget::sizeChanged(width, height);

    uiData->setSizeCallback(width, height);
}

#endif // DISTRHO_PLUGIN_HAS_EXTERNAL_UI
```

Here UIWidget is typedef'd to ExternalWindow in DistrhoUI.hpp, and uiData->setSizeCallback() results in the host calling effEditGetRect. Inside effEditGetRect:

```
                fVstRect.right  = fVstUI->getWidth();
                fVstRect.bottom = fVstUI->getHeight();
```

fVstUI->getWidth() and fVstUI->getHeight() both return 0.  Setting fVstRect.right and fVstRect.bottom to a constant also does not work, the plugin view just disappears.

I feel this is totally wrong and the fix is actually deeper lol. This issue and the plugin not showing on Live (maybe both things are related?) is the only thing that is preventing me from completely ditching DGL. Any hint?  thanks.

[...]

PR #319, opened 2021-09-03
Bugfix: EmbedExternalExampleUI resizing on Mac

lucianoiam:
Discussion at https://github.com/DISTRHO/DPF/issues/313

### DIFF
--- a/examples/EmbedExternalUI/EmbedExternalExampleUI.cpp
+++ b/examples/EmbedExternalUI/EmbedExternalExampleUI.cpp
@@ -93,14 +93,12 @@ public:
             [NSApp activateIgnoringOtherApps:YES];
         }
 
-        fView = [NSView new];
+        fView = [[NSView alloc] initWithFrame:NSMakeRect(0, 0, getWidth(), getHeight())];
         DISTRHO_SAFE_ASSERT_RETURN(fView != nullptr,);
 
-        [fView initWithFrame:NSMakeRect(0, 0, getWidth(), getHeight())];
-        [fView setAutoresizesSubviews:YES];
-        [fView setFrame:NSMakeRect(0, 0, getWidth(), getHeight())];
-        [fView setHidden:NO];
-        [fView setNeedsDisplay:YES];
+        fView.autoresizesSubviews = YES;
+        fView.wantsLayer = YES;
+        fView.layer.backgroundColor = NSColor.blueColor.CGColor;
 
         if (isEmbed())
         {
@@ -271,6 +269,9 @@ protected:
         UI::sizeChanged(width, height);
 
 #if defined(DISTRHO_OS_MAC)
+        NSRect rect = fView.frame;
+        rect.size = CGSizeMake((CGFloat)width, (CGFloat)height);
+        fView.frame = rect;
 #elif defined(DISTRHO_OS_WINDOWS)
 #else
         if (fWindow != 0)
@@ -341,6 +342,10 @@ protected:
 
     void uiIdle() override
     {
+        if (isEmbed()) {
+            return;
+        }
+
         // d_stdout("uiIdle");
 #if defined(DISTRHO_OS_MAC)
         NSAutoreleasePool* const pool = [[NSAutoreleasePool alloc] init];
--- a/examples/EmbedExternalUI/EmbedExternalExampleUI.cpp
+++ b/examples/EmbedExternalUI/EmbedExternalExampleUI.cpp
@@ -93,12 +93,13 @@ public:
             [NSApp activateIgnoringOtherApps:YES];
         }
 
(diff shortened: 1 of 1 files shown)


examples/EmbedExternalUI/EmbedExternalExampleUI.cpp
```diff
@@ -341,6 +342,10 @@ class EmbedExternalExampleUI : public UI
 
     void uiIdle() override
     {
+        if (isEmbed()) {
```
falkTX, line 345: this check must be only for macOS, on the other systems we need to poll to get our own custom events.

examples/EmbedExternalUI/EmbedExternalExampleUI.cpp
```diff
@@ -271,6 +269,9 @@ class EmbedExternalExampleUI : public UI
         UI::sizeChanged(width, height);
 
 #if defined(DISTRHO_OS_MAC)
+        NSRect rect = fView.frame;
```
falkTX, line 272: does this even do anything? you are not calling NSView functions which could trigger callbacks..

falkTX:
I do not quite understand this change tbh.
Why does initializing the view with a size change from allocating the view and then initializing the size?
It is all within the same function scope, so the host doesnt have access to the view until we are done with the whole thing..

examples/EmbedExternalUI/EmbedExternalExampleUI.cpp
lucianoiam: that is true, as discussed ...

falkTX: then I was not clear enough. the event loop is not needed on macos or windows, but it is always needed on linux since there is no native event loop to hook into.
on x11 we must always handle our own events.

examples/EmbedExternalUI/EmbedExternalExampleUI.cpp
lucianoiam: This is actually a call behind the scenes: 

` fView.frame = rect;`

It translates to

`[fView setFrame:rect];`

It is syntactic sugar, ObjC calls these "properties".

lucianoiam:
> I do not quite understand this change tbh.
> Why does initializing the view with a size vs/changes from allocating the view and then initializing the size?
> It is all within the same function scope, so the host doesnt have access to the view until we are done with the whole thing..

`[NSObject new]` translates to `[[NSObject alloc] init]`.  Then calling initWithFrame: would double-initialize the object. Also the original code was setting the frame twice, first by calling initWithFrame, then by calling setFrame.The standard way of initializing a view with a given size is `[[NSView alloc] initWithFrame:...]`.

`[NSView new]`  then `[view setFrame:]` or `view.frame = ...` would also be correct. I opted for the one-liner.

Worth noting it is fine to call methods (send messages, in ObjC parlance) in a null (nil) object (receiver).

Did I correctly understand ?

examples/EmbedExternalUI/EmbedExternalExampleUI.cpp
lucianoiam: nah you were clear enough. I agreed to that but then wrote that piece in the wrong place... now fixing that.

lucianoiam:
Last commit avoids ObjC properties, somehow making the intention of resizing the child window more explicit, it adheres more to the current ObjC code style. Also it restricts the isEmbed() check in uiIdle() to Mac ( will have a look at Windows later ... )

examples/EmbedExternalUI/EmbedExternalExampleUI.cpp
falkTX: ah I was not aware of this, thanks. ObjC is quite weird..

lucianoiam: Apple gradually started adding sugar after iOS was invented, if I correctly recall. The core OOP ideas come from Smalltalk, hence the weirdness...

falkTX:
GitHub is getting weird, not sure if I made a comment about the size or not..
Unless https://github.com/DISTRHO/DPF/pull/319/files is wrong, I do not see the view having an initial size.

lucianoiam:
This line does it

`[fView setFrame:NSMakeRect(0, 0, getWidth(), getHeight())];`

Hope I didn't make things unnecessarily confusing with the 2nd commit. I split the view init in two:

`[NSView new]`   like it was before, this is exactly the same as alloc, then init

then

`[fView setFrame:...]`

The frame contains size and position information. Either new, then setFrame is called, or alloc then initWithFrame.

falkTX:
wow, I must be blind.

well all seems fine, thanks!

PR #320, opened 2021-09-04
Fix build on Linux

lucianoiam:
Moving one DPF include after X11/Xresource.h fixes the issue.


lucianoiam:
Just tried compiling a fresh clone of 9538417 and it works, go figure.

falkTX:
External UI is not part of the default build though, so this error is not caught by default now.

PR #323, opened 2021-09-07
Bugfix: build error with DISTRHO_PLUGIN_HAS_EXTERNAL_UI

lucianoiam:
This is similar to https://github.com/DISTRHO/DPF/pull/320 except:

- Issue properly described, it was only affecting external UI not the whole framework
- Do it in a cleaner way than changing header include order

### DIFF
--- a/distrho/src/DistrhoUI.cpp
+++ b/distrho/src/DistrhoUI.cpp
@@ -25,10 +25,10 @@
 # endif
 #else
 # include "src/TopLevelWidgetPrivateData.hpp"
+# include "src/WindowPrivateData.hpp"
 #endif
 
 #include "DistrhoUIPrivateData.hpp"
-#include "src/WindowPrivateData.hpp"
 
 START_NAMESPACE_DISTRHO
 
--- a/distrho/src/DistrhoUI.cpp
+++ b/distrho/src/DistrhoUI.cpp
@@ -14,8 +14,8 @@
  * CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  */
 
-#include "DistrhoUIPrivateData.hpp"
-#include "src/WindowPrivateData.hpp"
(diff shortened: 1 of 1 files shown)


distrho/src/DistrhoUI.cpp
```diff
@@ -27,6 +27,9 @@
 # include "src/TopLevelWidgetPrivateData.hpp"
 #endif
 
+#include "DistrhoUIPrivateData.hpp"
+#include "src/WindowPrivateData.hpp"
```
falkTX, line 31: hmm this WindowPrivateData is for DGL stuff, so it better not be in use for external UIs.
Can you move it inside the ifdef after TopLevelWidgetPrivateData ?
If it doesnt break the usual build, it will be good for merging.
Thanks!

lucianoiam: Sure, done. Both main build and EmbedExternalUI build work.

falkTX: cool, thanks very much

lucianoiam:
Looks like CI is failing to build tests. `cd tests && make` works here, not sure that is the way to replicate locally.

```
../build/tests/Window.opengl 
30
assertion failure: "world != nullptr" in file ../dgl/src/ApplicationPrivateData.cpp, line 60
31
make: *** [Makefile:64: run] Segmentation fault (core dumped)
32
rm ../build/tests/Window.cpp.stub.o ../build/tests/Point.cpp.o ../build/tests/Window.cpp.cairo.o ../build/tests/Application.cpp.o ../build/tests/Color.cpp.o ../build/tests/Window.cpp.opengl.o
33
make: Leaving directory '/home/runner/work/DPF/DPF/tests'
34
Error: Process completed with exit code 2.

falkTX:
I think that is unrelated. restarted the build and the tests pass now.

PR #324, opened 2021-09-11
Prevent some compilation warnings on MinGW

lucianoiam:
Found this working solution https://github.com/official-stockfish/Stockfish/issues/1619#issuecomment-396776490 . There might be others...

Suppresses these warnings which are generated multiple times when compiling the examples:

```
Compiling DistrhoUIMain.cpp (JACK)
In file included from ../../distrho/DistrhoUIMain.cpp:17:
../../distrho/src/DistrhoUI.cpp: In function 'double DISTRHO::getDesktopScaleFactor(uintptr_t)':
../../distrho/src/DistrhoUI.cpp:62:15: warning: cast between incompatible function types from 'FARPROC' {aka 'long long int (*)()'} to 'PFN_GetProcessDpiAwareness' {aka 'long int (*)(void*, long unsigned int*)'} [-Wcast-function-type]
   62 |             = (PFN_GetProcessDpiAwareness)GetProcAddress(Shcore, "GetProcessDpiAwareness");
      |               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../../distrho/src/DistrhoUI.cpp:64:15: warning: cast between incompatible function types from 'FARPROC' {aka 'long long int (*)()'} to 'PFN_GetScaleFactorForMonitor' {aka 'long int (*)(HMONITOR__*, long unsigned int*)'} [-Wcast-function-type]
   64 |             = (PFN_GetScaleFactorForMonitor)GetProcAddress(Shcore, "GetScaleFactorForMonitor");
      |               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/dgl/src/pugl.cpp
+++ b/dgl/src/pugl.cpp
@@ -213,10 +213,17 @@ double puglGetDesktopScaleFactor(const PuglView* const view)
         typedef HRESULT(WINAPI* PFN_GetProcessDpiAwareness)(HANDLE, DWORD*);
         typedef HRESULT(WINAPI* PFN_GetScaleFactorForMonitor)(HMONITOR, DWORD*);
 
+# if defined(__GNUC__) && (__GNUC__ >= 9)
+#  pragma GCC diagnostic push
+#  pragma GCC diagnostic ignored "-Wcast-function-type"
+# endif
         const PFN_GetProcessDpiAwareness GetProcessDpiAwareness
             = (PFN_GetProcessDpiAwareness)GetProcAddress(Shcore, "GetProcessDpiAwareness");
         const PFN_GetScaleFactorForMonitor GetScaleFactorForMonitor
             = (PFN_GetScaleFactorForMonitor)GetProcAddress(Shcore, "GetScaleFactorForMonitor");
+# if defined(__GNUC__) && (__GNUC__ >= 9)
+#  pragma GCC diagnostic pop
+# endif
 
         DWORD dpiAware = 0;
         if (GetProcessDpiAwareness && GetScaleFactorForMonitor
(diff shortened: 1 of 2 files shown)


falkTX:
I saw this before, and this has been "fixed" in a few places in DPF.
tbh I dont quite get what the warning is all about, the function is casted as needed..

in any case, this double casting doesnt look nice.
if you look into NanoVG.cpp dpf already sets up some macros to ignore these warnings.
I think that is better vs the double casting you propose here.

lucianoiam:
Right, NanoVG.cpp contains a much cleaner solution:

```
# if defined(__GNUC__) && (__GNUC__ >= 9)
#  pragma GCC diagnostic push
#  pragma GCC diagnostic ignored "-Wcast-function-type"
# endif

...

# if defined(__GNUC__) && (__GNUC__ >= 9)
#  pragma GCC diagnostic pop
# endif
```

Force-pushed to replace first version

falkTX:
Thanks!
If you put the macros only around the 2 `GetProcAddress` calls it still silences the warnings, right?
That might look a bit cleaner.

lucianoiam:
Sure, I was unsure about the style. Fixed.

falkTX:
Style is always a tricky thing, I sorta like the way juce does it with indented macros, but it also gets confusing when lots of them as nested together..

By the way, don't we need this same change in pugl.cpp ?

falkTX:
Yeah I see in https://github.com/DISTRHO/DPF/blob/main/dgl/src/pugl.cpp#L216 we need to place the same macros.

lucianoiam:
That's correct, also updated!

falkTX:
Cool, thanks a lot!

PR #328, opened 2021-09-18
Exported symbol lists with CMake

jpcima:
Adds windows def files, `main` vst entrypoint, cmake support.


falkTX:
Thanks a lot!
So the version files are now only for non-mac/win, I will remove the win32 specific bits from there.

PR #338, opened 2021-10-09
Update NanoVG and stb-image sources, bring all VCV goodies too

falkTX:
@jpcima You have some plugins made with nanovg, can you give this a try?
Basically it is the updated version from VCVRack with the DPF custom additions on top.

For VCV itself this works, hopefully doesnt break anything else.

### DIFF
--- a/dgl/NanoVG.hpp
+++ b/dgl/NanoVG.hpp
@@ -943,6 +943,7 @@ private:
     inline void onDisplay() override
     {
         // NOTE maybe should use BaseWidget::getWindow().getScaleFactor() as 3rd arg ?
+        NanoVG::reset();
         NanoVG::beginFrame(BaseWidget::getWidth(), BaseWidget::getHeight());
         onNanoDisplay();
         NanoVG::endFrame();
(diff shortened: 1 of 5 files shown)


jpcima:
Not immediately, the spectacle-analyzer plugin wants a port to the newer model.
- change all to Subwidget
- mouse position handling, previously int and now double
- checking if my old widget hacks are still relevant in new DGL

falkTX:
> * mouse position handling, previously int and now double

about this one, there are a few cast operators in place to allow to use old code as-is.
if there is one missing let me know, perhaps we can add one more.

jpcima:
> But getting the old behaviour should be as simple as:

Rewriting in this way doesn't change anything.

falkTX:
I dont know then, but feel free to ping me if you need/want help with the update

I tested this branch with the nanovg stuff I have, everything seems to work just like before.
Going ahead and placing this on develop branch then.

jpcima:
> I dont know then, but feel free to ping me if you need/want help with the update

@falkTX it's the updated source which can't work properly on new DPF
https://github.com/jpcima/spectacle/pull/new/with-new-dpf

example: click "Scale" tool bar icon, and no other icon can get interaction from that point on
also I don't understand how the change corrected the event propagation supposedly.
What was wrong about propagating from child to parent? now it seems as if toplevel takes exception to this behavior.

falkTX:
Ok I will take a quick loook

PR #341, opened 2021-10-15
Allow to build VST3 for plugins with external UI

lucianoiam:
PR fixes this compilation issue:

```
In file included from ../../distrho/DistrhoUIMain.cpp:30:
../../distrho/src/DistrhoUIVST3.cpp: In member function ‘v3_result DISTRHO::UIVst3::onKeyDown(int16_t, int16_t, int16_t)’:
../../distrho/src/DistrhoUIVST3.cpp:222:20: error: ‘class DISTRHO::UIExporter’ has no member named ‘handlePluginKeyboardVST3’
  222 |         return fUI.handlePluginKeyboardVST3(true, static_cast<uint>(keychar), dglcode, dglmods) ? V3_TRUE : V3_FALSE;
      |                    ^~~~~~~~~~~~~~~~~~~~~~~~
../../distrho/src/DistrhoUIVST3.cpp: In member function ‘v3_result DISTRHO::UIVst3::onKeyUp(int16_t, int16_t, int16_t)’:
../../distrho/src/DistrhoUIVST3.cpp:253:20: error: ‘class DISTRHO::UIExporter’ has no member named ‘handlePluginKeyboardVST3’
  253 |         return fUI.handlePluginKeyboardVST3(false, static_cast<uint>(keychar), dglcode, dglmods) ? V3_TRUE : V3_FALSE;
      |                    ^~~~~~~~~~~~~~~~~~~~~~~~
make: *** [../../Makefile.plugins.mk:302: ../../build/d_embedextui/DistrhoUIMain_VST3.cpp.o] Error 1
```

### DIFF
--- a/distrho/src/DistrhoUIVST3.cpp
+++ b/distrho/src/DistrhoUIVST3.cpp
@@ -195,7 +195,7 @@ public:
     {
         d_stdout("onKeyDown %i %i %x\n", keychar, keycode, modifiers);
         DISTRHO_SAFE_ASSERT_INT_RETURN(keychar >= 0 && keychar < 0x7f, keychar, V3_FALSE);
-
+#if !DISTRHO_PLUGIN_HAS_EXTERNAL_UI
         using namespace DGL_NAMESPACE;
 
         // TODO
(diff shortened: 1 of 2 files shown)


falkTX:
Oh I forgot about this one, thanks!

But.. since the external UI won't receive those events, would be best to just ifdef the UIVST3 stuff entirely, and return not-implemented from the C call directly, skipping the C++ cast.

btw, sorta related to this...
as typical with stuff steinberg does, no one cares so I have yet to see a host that sends out these events :sweat_smile: 
this is why it is incomplete, I am unable to test the plugin part...

lucianoiam:
I am not sure to fully understand what is the root issue. In the case of the webview UI, it already seems to work on Mac/VST3. There are some bugs but the native Cocoa widget WKWebView renders and responds to user events as expected. Still need to spend time on the bugs. Naturally this is just a single case, likely missing something here...

`CADisplayLink`  might be useful for the idle timer according to https://stackoverflow.com/questions/30856330/trying-to-create-an-accurate-timer-with-gcd and https://dmtopolog.com/cadisplaylink-and-its-applications/

In the case of Windows/VST3 yes, the web view does not even render.  I am not really a Windows programmer so not sure where to look, but this sounds useful https://docs.microsoft.com/en-us/windows/win32/winmsg/window-features#message-only-windows . The webview project does add an invisible window with  `GWLP_USERDATA` to the host provided hwnd for some purposes.  Quite sure this is not new at all :)

Issue #359, opened 2022-01-09
VST3 + DISTRHO_PLUGIN_HAS_EXTERNAL_UI view sizing


lucianoiam:
Excellent work on the VST3 implementation !

I inserted a debug message into DistrhoUIVST3.cpp like this:

```
    void doIdleStuff()
    {
        d_stdout("doIdleStuff()");

     ...
```

And noticed the message is never printed when `DISTRHO_PLUGIN_HAS_EXTERNAL_UI` is defined.

Is this the expected behavior?  specifically my plugins are failing to request a new size, while the VST2 builds work as expected. DPF prints this instead:

`plugin->host setSize 600 300 (IGNORED, host resize active)`

Looks like the flag `fIsResizingFromPlugin` is always true and never cleared because `doIdleStuff()` is never called. There is a single `fIsResizingFromPlugin = false` line and it belongs to that method. I think this flag prevents a resizing feedback loop but honestly I haven't thoroughly studied this.

falkTX:
I didnt spend any time on vst3 and external ui yet, there is going to be a need to write some custom code to handle that properly on non-x11 platforms.
I would say the vst3 ui is not ready yet, and focus should be on finalizing the non-external stuff first. At least right now on there are issues reopening the UI on bitwig.

Once the non-external stuff works fine, I will see about the external case.

But if you want to help out, the situation here is that vst3 does not expose ui idle time for win/mac, only on x11 (by the use of the host-timer interface).
On win/mac the way I got this to work is to register a callback with pugl that will drive the plugin UI, rather than the host calling ui idle and that driving the pugl stuff (which is how all other formats work in dpf).
For x11 though, it should work similar to vst2 in theory, so it will be easier to make that work first.

lucianoiam:
Naive question but why VST3 cannot just work like VST2 in this regards? is there a fundamental difference, like VST2 requiring the host to call a method at a regular interval, and VST3 not?

I see `DPF_VST3_USING_HOST_RUN_LOOP` will only work for Linux but not sure why.

For Mac there is GCD or `NSTimer`. A further method optimized for graphics is called `CADisplayLink`. Just throwing in some names, maybe you already saw all this. Would need to better understand the stack myself.

Leaving some links found on the way
https://dmtopolog.com/cadisplaylink-and-its-applications/
https://stackoverflow.com/questions/55131532/difference-between-dispatchsourcetimer-timer-and-asyncafter

falkTX:
[...]

PR #360, opened 2022-01-11
Idle timer for external UI

lucianoiam:
Related to https://github.com/DISTRHO/DPF/issues/359

### DIFF
--- a/distrho/src/DistrhoUIVST3.cpp
+++ b/distrho/src/DistrhoUIVST3.cpp
@@ -21,6 +21,10 @@
 #include "travesty/host.h"
 #include "travesty/view.h"
 
+#if defined(DISTRHO_OS_WINDOWS)
+#include <winuser.h>
+#endif
+
 /* TODO items:
  * - mousewheel event
  * - key down/up events
@@ -127,6 +131,12 @@ public:
           fScaleFactor(scaleFactor),
           fIsResizingFromPlugin(false),
           fIsResizingFromHost(willResizeFromHost),
+#if defined(DISTRHO_OS_MAC)
+          fTimerPtr(nullptr),
+#endif
+#if defined(DISTRHO_OS_WINDOWS)
+          fTimerHwnd(reinterpret_cast<HWND>(winId)),
+#endif
           fUI(this, winId, sampleRate,
               editParameterCallback,
               setParameterCallback,
@@ -546,13 +556,16 @@ private:
     bool fIsResizingFromPlugin;
     bool fIsResizingFromHost;
 
-    // Plugin UI (after VST3 stuff so the UI can call into us during its constructor)
-    UIExporter fUI;
-
     // Native timer support
 #if defined(DISTRHO_OS_MAC)
-    CFRunLoopTimerRef fTimer;
+    CFRunLoopTimerRef fTimerPtr;
 #endif
+#if defined(DISTRHO_OS_WINDOWS)
+    HWND fTimerHwnd;
+#endif
+
+    // Plugin UI (after VST3 stuff so the UI can call into us during its constructor)
+    UIExporter fUI;
 
     // ----------------------------------------------------------------------------------------------------------------
     // helper functions called during message passing
@@ -733,15 +746,15 @@ private:
         const CFTimeInterval t = static_cast<double>(timerFrequencyInMs) / 1000.0;
         CFRunLoopTimerContext ctx = {};
         ctx.info = this;
-        fTimer = CFRunLoopTimerCreate(kCFAllocatorDefault, CFAbsoluteTimeGetCurrent() + t, t, 0, 0,
+        fTimerPtr = CFRunLoopTimerCreate(kCFAllocatorDefault, CFAbsoluteTimeGetCurrent() + t, t, 0, 0,
             UIVst3::nativeIdleTimerCallback, &ctx);
-        CFRunLoopAddTimer(CFRunLoopGetCurrent(), fTimer, kCFRunLoopCommonModes);
+        CFRunLoopAddTimer(CFRunLoopGetCurrent(), fTimerPtr, kCFRunLoopCommonModes);
     }
 
     void nativeIdleTimerDestroy()
     {
-        CFRunLoopRemoveTimer(CFRunLoopGetCurrent(), fTimer, kCFRunLoopCommonModes);
-        CFRelease(fTimer);
+        CFRunLoopRemoveTimer(CFRunLoopGetCurrent(), fTimerPtr, kCFRunLoopCommonModes);
+        CFRelease(fTimerPtr);
     }
 
     static void nativeIdleTimerCallback(CFRunLoopTimerRef timer, void *info)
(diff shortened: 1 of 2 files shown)


falkTX:
Very appreciated, but I wont merge until we have the win32 side too.
Mostly because, from my previous tests and attempts at this, doing this on win32 is much more difficult and is likely to need change a few core classes.

lucianoiam:
Makes sense, will try Win32 in the upcoming days.

lucianoiam:
Reminder, the Mac idle callback is being called on the main thread and not sure that is the correct thing to do.

lucianoiam:
Done for Win32. However this comment keeps me thinking if there's something missing:

>  doing this on win32 is much more difficult and is likely to need changing a few core classes.

The Windows implementation looks very similar to Mac's.

falkTX:
huh so you are using the timer id as a pointer?? hmm I didnt think think of that, somehow thought it was not possible.
you tested this to work?

my issue is that the HWND belonging to pugl was not able to let me access the WinProc so I couldnt process the timer messages myself.

distrho/src/DistrhoUIVST3.cpp
```diff
@@ -538,6 +556,14 @@ class UIVst3
     bool fIsResizingFromPlugin;
     bool fIsResizingFromHost;
 
+    // Native timer support
```
falkTX, line 559: these should be under a `#if DISTRHO_PLUGIN_HAS_EXTERNAL_UI` as they are unused otherwise

distrho/src/DistrhoUIVST3.cpp
```diff
@@ -168,8 +182,12 @@ class UIVst3
         if (fConnection != nullptr)
             connect(fConnection);
 
-#if !DISTRHO_PLUGIN_HAS_EXTERNAL_UI && (defined(DISTRHO_OS_MAC) || defined(DISTRHO_OS_WINDOWS))
+#if defined(DISTRHO_OS_MAC) || defined(DISTRHO_OS_WINDOWS)
+# if DISTRHO_PLUGIN_HAS_EXTERNAL_UI
+        nativeIdleTimerCreate(DPF_VST3_TIMER_INTERVAL);
```
falkTX, line 187: I think it makes sense to put all the idle stuff together, so move the line below `fUI.addIdleCallbackForVST3(this, DPF_VST3_TIMER_INTERVAL);` into this new function, way less confusing code that way

lucianoiam:
- CI build fails because of some casting issue, build works here (MinGW running on Windows). Do you know how to properly fix it?

- Stole the user data idea from here https://stackoverflow.com/questions/4625184/passing-user-data-with-settimer . It is not very clean but works... ideally GWLP_USERDATA should be used but I think the window is owned by pugl so there should be some additional coordination, I guess

- The callback runs as expected, tested by placing `d_stdout()` calls in `UIVst3::onTimer()`.  Honestly I don't understand the MSDN docs on this https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-settimer, does the callback function being called depend on processing WM_TIMER? or are both concepts independent from each other?  The hwnd is there because passing NULL hwnd to `SetTimer()` does not allow passing a custom event ID.

lucianoiam:
"When you specify a TimerProc callback function, the default window procedure calls the callback function when it processes WM_TIMER. "

If I correctly understand the hwnd UIVst3 receives is using the default window proc? ie,  it is not Pugl's, otherwise the callback would never be called. I don't know the window hierarchy...

falkTX:
> * CI build fails because of some casting issue, build works here (MinGW running on Windows). Do you know how to properly fix it?

you need a `WINAPI` macro for the static function.
or something like that.



> * does the callback function being called depend on processing WM_TIMER? or are both concepts independent from each other?  The hwnd is there because passing NULL hwnd to `SetTimer()` does not allow passing a custom event ID.

I think this only works if the HWND owner is not already processing timer ids, because afaik the SetTimer depends on it.
But it is not something I understand very well.

falkTX:
> If I correctly understand the hwnd UIVst3 receives is using the default window proc? ie, it is not Pugl's, otherwise the callback would never be called. I don't know the window hierarchy...

IIRC during my testing I was unable to rely on SetTimer due to pugl having its own way to handle timers.

lucianoiam:
Maybe it is safer to create a child invisible window for this?

lucianoiam:
> IIRC

Wait. Is Pugl being used for external UI?

falkTX:
> > IIRC
> 
> Wait. Is Pugl being used for external UI?

no, but initially I tried to set up timers without having to rely on pugl.
then gave up and just used it, ignoring the external-ui case for the moment.

but back we are again at it :grin:

lucianoiam:
Ok, thanks. Will work a little more on this I think we have something now.

distrho/src/DistrhoUIVST3.cpp
lucianoiam: Are you sure these are not used by the Pugl-based implementation?

distrho/src/DistrhoUIVST3.cpp
lucianoiam: Sorry didn't get this one, there is nothing below `fUI.addIdleCallbackForVST3(this, DPF_VST3_TIMER_INTERVAL);`. What should be moved into `nativeIdleTimerCreate()`?

lucianoiam:
I believe this is now complete, updated the Win32 part so it does not rely on the native parent window handle passed to UIVst3 ctor anymore when setting up a timer. It is a mistake to assume this external (to the plugin) window will process WM_TIMER messages like DefWindowProc does. FYI according to docs DefWindowProc calls the callback function passed to `SetTimer()` when processing WM_TIMER messages.

There might be some code style issues, left some comments above.

lucianoiam:
Also using a custom window allows using GWLP_USERDATA in place of the  eventId=this dirty trick

falkTX:
I would do some things slightly different, and try to cleanup the code so there is not a whole bunch of ifdefs very close to each other.
Will do this at a later point soon.

lucianoiam:
Ok, sure! thanks.

lucianoiam:
Just letting you know this code has been tested regularly for two months and it is working great, in case reformat/refactor and merge is decided.

Unrelated but still VST3, I have never been able to run DPF plugins compiled in that format on Live/macOS (one of the most popular non-free setups?). I know VST3 needs some polishing but perhaps you were not aware of this particular combination.

Not very useful screenshot below... need to debug.
Screen Shot 2022-03-13 at 19 21 32

falkTX:
Implemented this now in 4313bac8873d9f62c9a6319a4a8faba8219147e2, though I did not test it at all.
A little native loop helper class is in place there to ensure it all works more or less the same, be it regular UI or external.

Let me know if/how it works for you.
Will leave this PR open for now until you confirm it works.

falkTX:
on the macOS thing, I will handle it a bit later. will try to find another way to fix the include.

lucianoiam:
External UI timer works as expected on Mac and Windows. Did not test Linux since there is no need for the custom timer on that platform.

I think this PR can be closed, thanks!

falkTX:
Cool, closing then.
Thanks for checking

Issue #364, opened 2022-01-29
VST3: crash regression on Mac


lucianoiam:
FYI, I know VST3 is WIP.

Did not happen before bdf0759. Only affects Mac, plugins still work fine on Linux and Windows. Test host is always REAPER.

`fDummyAudioBuffer` seems null at this point:
https://github.com/DISTRHO/DPF/blob/a349e82a719d2e6de9394e211f355ca1632b97a7/distrho/src/DistrhoPluginVST3.cpp#L1156

```
(lldb) bt
* thread #12, name = 'com.apple.audio.IOThread.client', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
  * frame #0: 0x00007ff816c8c675 libsystem_platform.dylib`_platform_bzero$VARIANT$Haswell + 53
    frame #1: 0x000000011a3ace07 d_webgain`DISTRHO::PluginVst3::process(this=0x0000600003074a20, data=0x000070000e58ac80) at DistrhoPluginVST3.cpp:1149:9 [opt]
    frame #2: 0x000000010060e4c7 REAPER`VST3_ProcessReplacing(AEffect*, float**, float**, int) + 2823
    frame #3: 0x00000001003ed006 REAPER`VST_HostedPlugin::ProcessSamples(int, double*, int, int, int, double, midi_List*, bool*, double, double, double, bool, bool) + 10262
    frame #4: 0x0000000100415b4f REAPER`FxDsp::processFxDsp(int, double*, int, int, int, double, midi_List*, double, bool, double, double, double, double) + 4735
    frame #5: 0x000000010042e1d6 REAPER`FxChain::ProcessChain(int, double*, int, int, int, int, double, midi_List*, double, bool) + 1030
    frame #6: 0x00000001000ef6f9 REAPER`MediaTrack::RenderSamples_nocache(double, long long, double*, int, int, double, MediaTrack* const*, int, int*, bool, int, int, bool*, SyncSMP_Context*) + 22857
    frame #7: 0x00000001000f5f90 REAPER`MediaTrack::RenderSamples(double, long long, int, Track_RS_Output*, int, double, MediaTrack* const*, int, midi_List*, int, int, int, int, int, MediaTrack::Track_SendRec*, bool*, SyncSMP_Context*) + 864
[...]

PR #365, opened 2022-01-30
Fix issue #364

lucianoiam:
https://github.com/DISTRHO/DPF/issues/364

Deal with `fDummyAudioBuffer` lifecycle in a supposedly safer way.

### DIFF
--- a/distrho/src/DistrhoPluginVST3.cpp
+++ b/distrho/src/DistrhoPluginVST3.cpp
@@ -1071,9 +1071,6 @@ public:
         else
         {
             fPlugin.deactivateIfNeeded();
-
-            delete[] fDummyAudioBuffer;
-            fDummyAudioBuffer = nullptr;
         }
 
         return V3_OK;


distrho/src/DistrhoPluginVST3.cpp
```diff
@@ -1065,6 +1065,9 @@ class PluginVst3
     {
         if (processing)
         {
+            delete[] fDummyAudioBuffer;
```
falkTX, line 1068: I suggest to not even do this part at all.
Buffer is deleted on plugin destructor, should be fine then.

lucianoiam: That is correct, overlooked that.

falkTX: I was meaning the whole block actually.
dont recreate the buffer here, seems pointless

lucianoiam: Where should it be created? as detailed in the issue report, `setProcessing(false)` can be called before `process()` causing a crash. Maybe a better solution is reverting this? and:

1. Have `process()` check if `setProcessing(true)` was called and return early otherwise. We agree calling `setProcessing(false)` before `process()` does not make sense but the host is doing it.

2. Null check `fDummyAudioBuffer` before writing it during `process()` . Not sure of side effects.

falkTX: as I see it, but might be wrong, it should only be recreated during `setupProcessing`.
if the nominal buffer size changes at any point, the host is supposed to call that.

lucianoiam: Wait maybe I misunderstood the first comment here. Are you talking about leaving array init to `setupProcessing()`, strip `setProcessing()` of any memory mgmt, and leave deletion *only*  for the destructor (original version calls delete in two places).

In other words revert everything and remove the delete from `setProcessing(false)`.

falkTX: yes that is correct.
do you see any problems with that approach? it wastes a bit of memory, but it is comes as side effect from more VST3 weirdness

lucianoiam: Not at all, it is the simplest and also more defensive solution. Unfortunately the latter comes with a cost but not I'd say it's more a philosophical one. Will push shortly.

lucianoiam: FYI, this is ready and tested.

falkTX:
Thanks!

PR #367, opened 2022-02-11
Fix make clean for vst3

noisecode3:
I think It makes sense to clean up .vst3 the same way .lv2 gets clean up. Or maybe I missed something. By the way thanks for vst3 support and all the work on DPF.

### DIFF
--- a/Makefile.plugins.mk
+++ b/Makefile.plugins.mk
@@ -316,7 +316,7 @@ $(BUILD_DIR)/%.mm.o: %.mm
 
 clean:
 	rm -rf $(BUILD_DIR)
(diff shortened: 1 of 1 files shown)


falkTX:
ah yes, you are quite correct.
can you also add `$(TARGET_DIR)/$(NAME).vst` while at it too? there is a flag now to turn the vst single file into a bundle, so we can put resource files into it.
thanks!

noisecode3:
Its added :)

lucianoiam:
I was about to open an identical PR, so +1 to this

falkTX:
Handled in e2dcae3758684b8c6a3157da71aad3757c9e652c
Thanks

PR #375, opened 2022-05-01
Update travesty

jpcima:


distrho/src/travesty/unit.h
```diff
@@ -0,0 +1,72 @@
...
+ * CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+#pragma once
+
+#include "base.h"
+
+#include "align_push.h"
+
+struct v3_unit_info
```
falkTX, line 23: the opening brack must be in the same line, alike the rest of the code

jpcima: Sure, that is fixed

PR #389, opened 2022-11-24
Vst3 io fixes

x42:


falkTX:
First one makes sense, I did the same in CLAP but forgot about it for VST3.

I am not so certain about the 2nd one though. JUCE hosts are particularly picky on what plugin sets as bus arrangement, easily bailing out if the bitmask is from anything not known.
From what I understand of your patch, it would allow to change a stereo bus into a mono one, and this is not supported in DPF. It is quite intentional that DPF does not allow to change the speaker arrangement of a bus, only turning it on or off and nothing else.

falkTX:
1st patch merged to develop as 36f018dacdcb96a7e858f2c078e6b876b1921119 thanks

x42:
DPF handles this just fine. DistrhoPluginVST3.cpp `process` uses `fDummyAudioBuffer` for disabled ports.

(NB use different buffers for input and output: silence for inputs, scratch buffers for outputs. Currently writing to the shared fDummyAudioBuffer modifies input data. The current approach only works if input data is read before any outputs is written).

falkTX:
"fine" depends on the perspective.

DPF plugins never know if host changes something so a stereo signal suddenly becomes mono, or vice-versa.
For DPF the idea is that audio ports are static. What becomes optional is "disabling" a particular group of ports, but then the entire group is either on or off, their actual port count must not change.

The point of input vs output buffer reusage is a good one though

x42:
The 2nd patch is also not critical. IIUC VST3 host needs to provide a valid buffers if a bus is enabled, regardless of the speaker arrangement.

falkTX:
Does Ardour provide a way to disable individual VST3 buses? I could give this some testing if so, my comment in the code regarding disabling (or not) certain buses is because I couldn't find a simple way to test it.
DPF should allow to at least disable certain buses, then providing the dummy buffer to the ports related to disabled buses.

If you tell me how to test/verify this via Ardour, I can give it spin soon-ish.

x42:
> Does Ardour provide a way to disable individual VST3 buses? 

Yes, one can use pin-connections to dis/connect individual ports.  If all ports of a bus are disconnected, the bus is disabled.
For VST3 however this is still work-in-progress (not yet in git/master), in current git this only works for the first main bus and first aux bus with VST3.

falkTX:
> PS. I am using Cardinal to test ardour's VST multi-bus support. and disabling busses altogether `Vst::IComponent::activateBus` does work.

You mean without the 2nd patch this already works as expected? I didnt actually try this yet...

btw, once this stabilizes on ardour side, give me a ping.
I am interested on doing some testing to ensure DPF side is also correct for a variety of cases.

x42:
> You mean without the 2nd patch this already works as expected?

Yes, a bus can be disabled completely (`activateBus(..., busID, false);`), which correctly sets DPF's `fEnabledInputs[i] = false` for all ports matching the busID,  and then fDummyAudioBuffer is used for those ports.

PR #394, opened 2022-12-13
CMake: Add external UI support

AnClark:
This PR will allow building external UI with CMake. Simply set `UI_TYPE` to `external` when using `dpf_add_plugin()`.

I've also added `CMakeLists.txt` for these two example plugins which shows how to use external UI:
- ExternalUI
- EmbedExternalUI

NOTE: Only tested on Arch Linux.


falkTX:
looks good, though I dont understand why `_dgl_library`needs to be set here

AnClark:
> I dont understand why _dgl_libraryneeds to be set here

This is because of the function `dpf__add_ui_main()`.

By the contributor's design, it has a parameter `HAS_UI`. The function is called in `dpf__build_<PLUGIN_TYPE>()` , and he passes `_dgl_library` as `HAS_UI`, just like this:

https://github.com/DISTRHO/DPF/blob/22413340a6d8ef2ffbf38ce841fb44c448a1a84a/cmake/DPF-plugin.cmake#L369

falkTX:
I think it makes more sense to have the call to `dpf__add_ui_main` have some general logic of "dgl-library or external-ui".
The `DGL_LIBRARY` var makes sense as it is, contains the lib property. So the check/call just needs some tweaking.

Is that doable? My knowledge of cmake is quite limited, dont actually know the syntax to do this.. So contributions here are very appreciated.

AnClark:
Yes, it's doable. Patch e392f065 has reduced ambiguity of `_dgl_library`, and use a more common vairalbe `_dgl_has_ui` instead.

AnClark:
Yes, it's doable. Patch e392f065 has reduced ambiguity of `_dgl_library`, and use a more common vairalbe `_dgl_has_ui` instead.

AnClark:
Hi falkTX! Is it possible to merge this PR? Now everything works well.

falkTX:
Merged in 189973dd2ba18912d4bd43af966108051deebd49, thanks!

PS: Please use the develop branch for pull requests next time

AnClark:
Also thanks a lot! Next time I will contribute to the develop branch.

AnClark:
Also thanks a lot! Next time I will contribute to the develop branch.

The only regret is that the squashed commit loses contributor information. Maybe I should merge any possible commits into 1 commit before submitting a PR.

AnClark:
Also thanks a lot! Next time I will contribute to the develop branch.

The only regret is that the squashed commit loses contributor information. Maybe I should merge any possible commits into 1 commit before submitting a PR.

falkTX:
> The only regret is that the squashed commit loses contributor information. Maybe I should merge any possible commits into 1 commit before submitting a PR.

yeah I had to do it via command-line manually, but couldnt figure out the proper way to handle this.

AnClark:
> yeah I had to do it via command-line manually, but couldnt figure out the proper way to handle this.

Maybe here's a solution. Use `git rebase`, then amend that commit with:

```
git commit --amend --author "AnClark Liu <anclarkliu@outlook.com>" --no-edit
```

falkTX:
right, that works. had to force-push and cherry-pick the commits after it, but fine as it was just 3

AnClark:
> right, that works. had to force-push and cherry-pick the commits after it, but fine as it was just 3

Thanks a lot!

Issue #395, opened 2022-12-31
DPF fails to build when using Clang?


pdesaulniers:
If I try to compile Wolf Shaper on Arch Linux with the following command:

```
CC=clang CXX=clang++ make
```

...I get the error `unknown argument: '-fno-gnu-unique'`:

```
CC=clang CXX=clang++ make
grep: warning: stray \ before #
make -C dpf/dgl opengl
make[1]: Entering directory '/home/patrick/Code/wolf-shaper/dpf/dgl'
grep: warning: stray \ before #
Compiling src/Application.cpp
clang-14: error: unknown argument: '-fno-gnu-unique'
make[1]: *** [Makefile:181: ../build/dgl/Application.cpp.o] Error 1
make[1]: Leaving directory '/home/patrick/Code/wolf-shaper/dpf/dgl'
make: *** [Makefile:18: libs] Error 2
```

It seems like Clang is not being auto-detected correctly, which causes this branch to be executed: https://github.com/DISTRHO/DPF/blob/22413340a6d8ef2ffbf38ce841fb44c448a1a84a/Makefile.base.mk#L220-L222

On my system, the value of `TARGET_COMPILER` is an empty string.

Any clues? Do I need to pass `CLANG=true` when building the plugin?

falkTX:
This is set in https://github.com/DISTRHO/DPF/blob/main/Makefile.base.mk#L38

It has worked for me in linux gcc, linux mingw, linux clang and macos clang.

What seems to be different for you is makefile, are you running arch or something that would get something super fresh/new?

pdesaulniers:
Yeah, I believe this occurs due to the way `echo` behaves on my system. This line does not return anything:

```
echo '#ifdef __clang__\nclang\n#else\ngcc\n#endif' | clang -E -P -x c - 2>/dev/null
```

I need to use `echo -e` instead:

```
$ echo -e '#ifdef __clang__\nclang\n#else\ngcc\n#endif' | clang -E -P -x c - 2>/dev/null
                                                                                                                
clang 
```

Also, this line emits a warning:

```
$ echo '\#test' | grep -- '\#test'           
grep: warning: stray \ before #                                       
\#test
```

To avoid these issues, would it be ok to use the `printf` command instead of `echo`? AFAIK, the `printf` command should be fairly portable:

```
printf '#ifdef __clang__\nclang\n#else\ngcc\n#endif' | clang -E -P -x c - 2>/dev/null

clang                     
```

[...]

PR #397, opened 2023-01-01
Fix target compiler detection

pdesaulniers:
Suggested fix for https://github.com/DISTRHO/DPF/issues/395

### DIFF
--- a/Makefile.base.mk
+++ b/Makefile.base.mk
@@ -35,12 +35,10 @@ DPF_MAKEFILE_BASE_INCLUDED = true
 # ---------------------------------------------------------------------------------------------------------------------
 # Auto-detect target compiler if not defined
 
-TARGET_COMPILER = $(shell printf '#ifdef __clang__\nclang\n#else\ngcc\n#endif' | $(CC) -E -P -x c - 2>/dev/null)
-
 ifneq ($(CLANG),true)
 ifneq ($(GCC),true)
 
-ifneq (,$(findstring clang,$(TARGET_COMPILER)))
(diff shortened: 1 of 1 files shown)


falkTX:
See CI, this already fails

falkTX:
This works here:
```sh
TARGET_COMPILER = $(shell printf '\#ifdef __clang__\nclang\n\#else\ngcc\n\#endif' | $(CC) -E -P -x c - 2>/dev/null)
```

but I guess it will fail for you..

pdesaulniers:
Oh yes, sorry, I thought the hash sign escaping quirks were a part of `echo` :(

> but I guess it will fail for you..

Yes, that's right, it fails on my system.

Would it be reasonable to simplify the check to `ifneq (,$(findstring clang,$(CC)))` ?

falkTX:
> Would it be reasonable to simplify the check to `ifneq (,$(findstring clang,$(CC)))` ?

no, calling "gcc" on macos will run clang
the command "gcc" is often an alias to the real compiler.

falkTX:
the fix involves this exact check:
```
ifeq ($(shell echo '\#test' | grep -- '\#test'),\#test)
```
this is what detects if the makefile needs special escaping or not.

pdesaulniers:
The issue has been fixed by https://github.com/DISTRHO/DPF/commit/c7e4ac43f5228cee69b2453d95e56859d7912a50 and https://github.com/DISTRHO/DPF/commit/34952157be7db5e01c58266431aab253b59f9259, so I am closing this PR.

Issue #402, opened 2023-01-05
Link fails when using UI_TYPE external


thirtythreeforty:
I'm using current tip of `develop` (54e1fb847d046f97b23b94d9c002efe29b3a4240). I'm attempting to integrate #394 into my project, I added the line to my add_plugin call:

```cmake
dpf_add_plugin(MyPlugin
  TARGETS vst3
  UI_TYPE external
  FILES_DSP
      MyPlugin.cpp
  FILES_UI
      MyUI.cpp)
```

but the build fails with:

```
/usr/bin/ld: cannot find -ldgl-system-libs-definitions: No such file or directory
```

I think a call to `dpf__add_dgl_system_libs` is missing in this code path. Perhaps it should be added here?

```patch
diff --git a/cmake/DPF-plugin.cmake b/cmake/DPF-plugin.cmake
index fcba2e12..2c2ca6b1 100644
--- a/cmake/DPF-plugin.cmake
+++ b/cmake/DPF-plugin.cmake
@@ -116,6 +116,7 @@ function(dpf_add_plugin NAME)
       dpf__add_dgl_opengl("${_dpf_plugin_NO_SHARED_RESOURCES}")
       set(_dgl_library dgl-opengl)
     elseif(_dpf_plugin_UI_TYPE STREQUAL "external")
+      dpf__add_dgl_system_libs()
       set(_dgl_external ON)
     else()
       message(FATAL_ERROR "Unrecognized UI type for plugin: ${_dpf_plugin_UI_TYPE}")
```

I think it is working in the examples because _all_ the examples are in one big CMake script, so this path gets called eventually.

thirtythreeforty:
@AnClark have you successfully used this as a submodule in another project?

falkTX:
Seems like I need to create some out-of-dpf external ui plugin of sorts, to test this behaviour

AnClark:
> @AnClark have you successfully used this as a submodule in another project?

Not yet. I'm porting SynthV1 to DPF, which has a Qt-based UI, but still work in progress (not ready for compilation).

As far as I try, the EmbedExternalUI example in develop branch can successfully compile.

AnClark:
> I think a call to dpf__add_dgl_system_libs is missing in this code path. Perhaps it should be added here?

External UI does not really depend on DGL. So I didn't add `dpf__add_dgl_system_libs()` there.

AnClark:
@thirtythreeforty My pleasure!

PR #404, opened 2023-01-05
CMake: Fix dpf__add_dgl_system_libs link error on external UI build

AnClark:
Fix issue mentioned in #402.


falkTX:
That makes sense, thanks!

AnClark:
My pleasure!

PR #405, opened 2023-01-10
Fix loop variable error on LV2export

AnClark:
Location: "Write preset.ttl" section. Index of program was erroneously passed to plugin.getStateHints(), which made LV2export misbehave.


falkTX:
oh nice one. must have been copy&paste lapse on my side. thanks!

AnClark:
My pleasure!

PR #421, opened 2023-05-14
Fix typos in package-osx-bundles.sh

pdesaulniers:
On macOS, plugin files were missing from the .dmg file. It seems like this was caused by some typos in the `package-osx-bundles.sh` script. This PR appears to fix the issue.


falkTX:
Yep, that is one silly copy&paste lapse.. :sweat_smile: 
Thanks for the fix!

btw, some CI builds might fail as GitHub updated their Ubuntu image and broke things once again.

PR #432, opened 2023-09-28
Fix crash on CLAP hosts that do not implement latency extension

messmerd:
When DISTRHO_PLUGIN_WANT_LATENCY == 1, CLAP plugins assume the host implements the latency CLAP extension and do not check whether the pointer to the host's latency extension is nullptr. As a result, a null pointer dereference can occur when the host proceeds to use the plugin.

This PR makes plugins that require the latency extension to check that hosts actually implement it, and if this requirement is not met, plugin initialization will fail.


falkTX:
I dont think this is the right approach, plugins with latency need to be able to load in hosts that do not care to do latency compensation.
So we should switch the fix to just catch the null pointer before using it, making the host feature optional.

falkTX:
Fixed in 4f11f8cc49b24ede1735a16606e7bad5a52ab41d

Issue #441, opened 2023-12-07
CLAP: add option to set supported dialect (e.g. CLAP_NOTE_DIALECT_MIDI_MPE)


jfrey-xx:
Hello,

Fist of all thank you for this framework, I'm creating my first audio plugins with it, and it's a breeze to get working my various synths / effects / utilities, with so many formats at once :)

I've been testing my plugins with the newly released Bitwig 5.1 (in Demo mode) and at the same time I wanted to try some things with pitch bend. I realized first that with 5.1 Bitwig now seems to acknowledge the `CLAP_NOTE_DIALECT_MIDI` flag, e.g. the "note on" events are not anymore processed through `CLAP_EVENT_NOTE_ON` in [DistrhoPluginCLAP.cpp](https://github.com/DISTRHO/DPF/blob/63dfb7610bc37dee69f4a303f3e3362529d95f24/distrho/src/DistrhoPluginCLAP.cpp#L959), but as regular MIDI events. The current workaround might probably stick for other cases. (I'll just refer to https://github.com/DISTRHO/DPF/issues/383#issuecomment-1256610736 if that can help to sort things).

The real issue I faced is that with 5.1 the pitch bend is not sent by bitwig anymore to my clap plugins (e.g. using the "fine" knob of the "note transpose" Note FX). For some reason I had to change at least the "supported" dialect flag in `DistrhoPluginCLAP.cpp`, from `CLAP_NOTE_DIALECT_MIDI` to `CLAP_NOTE_DIALECT_MIDI_MPE`. I guess this follows (odd?) Bitwig's behavior for the other formats (VST2 & VST3), where pitch bend is only sent if "MPE" is selected in the inspector. At my level I don't see any side effects, I would think that letting people choose the dialect flags offer more features to developers but I don't have the full picture.

To have the pitch bend perfectly usable in Bitwig I should also find a way to enable the selection of the pitch bend range in the DAW. There is no such option in inspector as compared to the VST plugins, and the host seems to default to +/- 48 semitones. I will handle that with a parameter at the plugin level, a good idea anyhow.

I spoke only of pitch bend, there might be other parameters (aftertouch?) that might be enabled in Bitwig through this MPE flag, I can do more testing if necessary.

jfrey-xx:
The main issue is fixed now that #443 is merged, thanks!

PR #443, opened 2023-12-09
CLAP: add macro to use MIDI MPE dialect

jfrey-xx:
This is related to #441 

In the end I could not help but try a little something. At first I wanted to let people select any of the different dialects, but I was not sure how to properly retrieve the definitions in DPF (include the corresponding CLAP header or duplicate its `#define` somewhere? Neither seemed pretty). So I settled with just the "MPE" option for now, especially considering that the other dialects (MDI 2 or CLAP) might need to be properly parsed to reach the plugin.

Then I started to add flags to let user configure MPE for input or output separately, to declare it as supported only (and not "preferred")... but that started to look bloated without other use cases.

Hence the trivial modification. I still hope that can be of some help. I tried my best to mimic how the flags are set. Tell me if I can improve something.


falkTX:
While I agree with the base idea, I think making this clap specific is wrong.
Best to me seems to have a generic hint (0/off by default if unset) to report a plugin as supporting MPE. Then DPF takes care that the host knows about this, which is already covered here for CLAP.
The value is only a hint anyway, so it is fine to be generic.

falkTX:
While I agree with the base idea, I think making this clap specific is wrong.
Best to me seems to have a generic hint (0/off by default if unset) to report a plugin as supporting MPE. Then DPF takes care that the host knows about this, which is already covered here for CLAP.
The value is only a hint anyway, so it is fine to be generic.

falkTX:
also please make the PR target develop branch, not main.
new commits should always go into develop first. thanks

falkTX:
actually I think the best is to have default value as "undefined", where we do nothing special if unset.
but plugins can manually specify through 0 or 1 value that they support MPE (or not)

most plugin specs dont have a way to report this property anyhow

jfrey-xx:
Thanks for the feedback. I rebased on develop and used a generic name for the macro -- maybe there is a better name for that (`DISTRHO_PLUGIN_USE_MIDI_MPE ` ?), maybe you wanted no default value in `DistrhoInfo.hpp`.

falkTX:
sorry it took a while to get to this, I want to add a few little things on top but never had the time for it, so best to merge it as-is.
I just now need to handle it similar to the other flags

thanks!

jfrey-xx:
No worries, thank you for getting back to that :) Now I definitely have to polish and publish the plugins I have been working on!

falkTX:
> No worries, thank you for getting back to that :) Now I definitely have to polish and publish the plugins I have been working on!

yes! then please update https://github.com/DISTRHO/DPF/wiki/Plugins-made-with-DPF to mention those too

jfrey-xx:
Sure will!

PR #448, opened 2024-02-09
Use correct VST 3 directory name on PowerPC

taylordotfish:
The [VST 3 specification](https://steinbergmedia.github.io/vst3_dev_portal/pages/Technical+Documentation/Locations+Format/Plugin+Format.html#for-the-linux-platform) declares that, on Linux, the name of the directory containing the plugin binary should begin with `uname -m`, but Makefile.base.mk uses the output of `gcc -dumpmachine`, which isn't always equivalent. In particular, this is the case for PowerPC—for example, on 64-bit little-endian PowerPC, `uname -m` is `ppc64le`, but `gcc -dumpmachine` is `powerpc64le-linux-gnu`, so the binary directory is incorrectly named `powerpc64le-linux` rather than `ppc64le-linux`.

This PR fixes the directory name on PowerPC, and also adds support for PowerPC in the CMake version, which currently errors.


falkTX:
I was thinking of using `uname -m`  but then the issue with `uname -m` is that it reports the host architecture, where we could be cross-compiling for something else entirely.
So actually your solution works better.
Thanks!

PR #453, opened 2024-06-14
[WIP] VST3: Fix incorrect MIDI input values for CC and Note On/Off

Simon-L:
Here's an approach to fixing the issue. I couldn't figure out where the normalized value comes from but the issue is probably there because for a received value of 1, it's mapped to 0.007874 which is 1.0/127 and not 1.0/128 as it should be, at least from what I can gather.

https://github.com/DISTRHO/DPF/compare/main...Simon-L:DPF:fix-vst3-midi-values?expand=1#diff-74e172a4a71aa46c34a85756971124027fc8807d094f6cf7f339335cca676234R1994 This line looked suspicious as well but it doesn't seem to be executed at any moment when receiving CC or Note On/Off

Given the current normalized input, mutliply by 127 would give wrong result for some input values, dividing by 0.0078125, which is 1.0/128, fixes this.

### DIFF
--- a/distrho/src/DistrhoPluginVST3.cpp
+++ b/distrho/src/DistrhoPluginVST3.cpp
@@ -359,14 +359,14 @@ class PluginVst3
                     midiEvent.size = 3;
                     midiEvent.data[0] = 0x90 | (eventStorage.noteOn.channel & 0xf);
                     midiEvent.data[1] = eventStorage.noteOn.pitch;
-                    midiEvent.data[2] = std::max(0, std::min(127, (int)(eventStorage.noteOn.velocity / 0.0078125)));
+                    midiEvent.data[2] = std::max(0, std::min(127, (int)std::round(eventStorage.noteOn.velocity * 127)));
                     midiEvent.data[3] = 0;
                     break;
                 case NoteOff:
                     midiEvent.size = 3;
                     midiEvent.data[0] = 0x80 | (eventStorage.noteOff.channel & 0xf);
                     midiEvent.data[1] = eventStorage.noteOff.pitch;
-                    midiEvent.data[2] = std::max(0, std::min(127, (int)(eventStorage.noteOff.velocity / 0.0078125)));
+                    midiEvent.data[2] = std::max(0, std::min(127, (int)std::round(eventStorage.noteOff.velocity * 127)));
                     midiEvent.data[3] = 0;
                     break;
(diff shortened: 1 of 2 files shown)


Simon-L:
I did dive a bit into travesty as well but couldn't figure out much of the flow.

At least now I understand what you meant what you said MIDI is "faked" in VST3... :(

distrho/src/DistrhoPluginVST3.cpp
```diff
@@ -359,14 +359,14 @@ class PluginVst3
                     midiEvent.size = 3;
                     midiEvent.data[0] = 0x90 | (eventStorage.noteOn.channel & 0xf);
                     midiEvent.data[1] = eventStorage.noteOn.pitch;
-                    midiEvent.data[2] = std::max(0, std::min(127, (int)(eventStorage.noteOn.velocity * 127)));
+                    midiEvent.data[2] = std::max(0, std::min(127, (int)(eventStorage.noteOn.velocity / 0.0078125)));
```
falkTX, line 362: dont use magic values please, why is that better than * 127 ?

Simon-L: because for a CC value of 1, normalized is 0.007874 and `(int)(0.007874 * 127)` is 0, it's just a rounding issue so `std::round` works too and is indeed better.  
`(int)std::round(eventStorage.noteOn.velocity * 127)`

Simon-L:
Thanks for the comment, branch updated to using std::round. Tested with cc and note on/off.

I can't test Polypressure and Channelpressure cases but I think we can assume they're affected as well.

falkTX:
huh I didnt expect that to be the issue.
but then please replace the lines from this style:
```
(int)std::round(eventStorage.noteOn.velocity * 127)
```

using this instead:
```
d_roundToIntPositive(eventStorage.noteOn.velocity * 127)
```

I already have some little dpf functions to deal with cases like these, so lets use them.

let me know if that still works as intended.

falkTX:
@Simon-L are you able to check this?

Simon-L:
Sorry for the delay, yes and I can confirm it fixes it.

I have added the change as a new commit as well as reverting the prints I added in the MidiThrough example

Unless I overlooked something, this should be ready to merge!

falkTX:
looks good, thanks!

PR #456, opened 2024-06-17
add double click event to ImageBaseKnob

chrisherb:
Since a double click event was already implemented in the backend, I simply added it to the ImageBaseKnob for easy subscription in the plugin's UI class. Useful for resetting a knob to its default value for example.


falkTX:
makes sense, but please push against develop branch instead of main. thanks!

chrisherb:
done!

PR #457, opened 2024-06-18
add default value getter

chrisherb:
Add a getter to retrieve a knob's default value. Useful when having to reset a knob to its default state.


falkTX:
function needs "const" attribute, also please put it just before the `setDefault` call
thanks

falkTX:
thanks!

PR #463, opened 2024-09-15
Add khronos to gitignore in order to avoid "modified" git status

sevenc-nanashi:
I use submodule to include DPF to my project.
DPF clones khronos in its directory, but it causes "modified" status on git.
This PR adds khronos to gitignore to avoid that status.


sevenc-nanashi:
I'm recreating the branch to make it go to develop branch

PR #470, opened 2024-10-04
 Add khronos to gitignore in order to avoid "modified" git status

sevenc-nanashi:
I use submodule to include DPF to my project.
DPF clones khronos in its directory, but it causes "modified" status on git.
This PR adds khronos to gitignore to avoid that status.

This is the same as #463 expect the base branch.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,5 @@ build/
 docs/
 utils/lv2_ttl_generator
 utils/lv2_ttl_generator.dSYM/
+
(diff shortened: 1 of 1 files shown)


sevenc-nanashi:
Could you review this please?

dromer:
@sevenc-nanashi you modified pugl somehow, which should not be in this PR

sevenc-nanashi:
I fixed that.

PR #473, opened 2024-10-18
Reintroduce std::string_view constructor for String

thirtythreeforty:
This reintroduces the `std::string_view` constructor for String which was removed in 01aca7649c1a3a5ee20a47c5ecd3cb2e29395f89. I agree that the constexpr portion of the functionality was largely broken, so this reintroduction does not declare anything constexpr.

### DIFF
--- a/distrho/extra/String.hpp
+++ b/distrho/extra/String.hpp
@@ -22,6 +22,10 @@
 
 #include <algorithm>
 
+#if __cplusplus >= 201703L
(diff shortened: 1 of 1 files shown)


falkTX:
that approach doesnt really work without the constexpr details, because then the string view can go out of scope and the string would contain garbage data.

I added 34435e0d86fb97e0643fd997526549b064de2d91 which copies the string view contents

PR #493, opened 2025-03-17
Add update() to NanoImage

Simon-L:
Following suggestions on adding this, I added an update() method to the NanoImage class.

Inspired from: https://github.com/memononen/nanovg/issues/269#issuecomment-189381540
Possible example usage in the Meters UI example: https://gist.github.com/Simon-L/65ac25c8caddebf11190c70b9dfea8a7#file-exampleuimeters-cpp-L226-L252

### DIFF
--- a/dgl/NanoVG.hpp
+++ b/dgl/NanoVG.hpp
@@ -107,6 +107,7 @@ public:
     */
     Size<uint> getSize() const noexcept;
 
+    void update(const unsigned char* data);
    /**
       Get the OpenGL texture handle.
     */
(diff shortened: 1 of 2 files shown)


Simon-L:
This PR causes an issue somewhere that breaks the texture update in certain circumstances. One case found for now is when adding the NanoButton class found in the FileHandling example
Here's a minimal example reproducing the issue: https://github.com/Simon-L/DPF-minimal-nanoimage-update

Simon-L:
Thank you so much!

PR #498, opened 2025-04-02
Updated setState on AU to allow arbitrary key-value pairs on macOS

### LINKED ISSUE #494: AU setState always triggers  assert "it != fStateKeys.end()" (both develop and master)
I am currently working on release builds for one of our plugins (working fine on VST3, Clap and Standalone), but having some issues with AU, in particular the setState function always asserting for some reason.
is there anything specific in regards to setup that i am missing from the plugin side?

[dpf] assertion failure: "it != fStateKeys.end()" in file /redacted/build/_deps/dpf-src/distrho/src/DistrhoUIAU.mm, line 326

Other than that the rest of the functionality seems to work fine, but i can't use setState which is important for part of our plugin communication.
Any places i should start looking?

Thanks

TheOnlyJoey:
Should fix #494 

Confirmed working with the States plugin example on macOS 15.3 using Logic

### DIFF
--- a/distrho/src/DistrhoPluginAU.cpp
+++ b/distrho/src/DistrhoPluginAU.cpp
@@ -1651,25 +1651,23 @@ public:
            #if DISTRHO_PLUGIN_WANT_STATE
             DISTRHO_SAFE_ASSERT_UINT_RETURN(inElement < fStateCount, inElement, kAudioUnitErr_InvalidElement);
             {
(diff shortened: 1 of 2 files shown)


falkTX:
thanks! the coding style is quite different from existing one on these files, I will clean it up a bit

falkTX:
Merged as part of 3f1e740f7659c01cb29471bb314bf14fa6ccb300 thanks!

PR #501, opened 2025-05-15
valgrind: Add malloc / memalign suppression for real 1d fftw

zamaudio:
Because: Testing on actual CI showed different results than locally. See https://github.com/zamaudio/zam-plugins/actions/runs/15036703697/job/42259708310


zamaudio:
Updated PawPaw hash and got this as well: https://github.com/zamaudio/zam-plugins/actions/runs/15079537628/job/42393787764 which confirms this fix should work.

PR #506, opened 2025-09-07
Fixed USE_FILEBROWSER typo in makefile

isivisi:
USE_FILE_BROWSER wasn't working for me with save dialogs on linux. Changing this allowed the save dialog to work, the dbus implementation seems to have the working save dialog implementation unlike the x11 route.

Looks like setting USE_FILE_BROWSER doesn't enable the correct dbus flags because its checking for USE_FILEBROWSER instead

Your makefile also implies this was a typo
```
ifeq ($(USE_FILEBROWSER),true)
$(error typo detected use USE_FILE_BROWSER instead of USE_FILEBROWSER)
endif
```

### DIFF
--- a/Makefile.base.mk
+++ b/Makefile.base.mk
@@ -484,7 +484,7 @@ endif
 
 else
 
-ifeq ($(USE_FILEBROWSER)$(HAVE_DBUS),truetrue)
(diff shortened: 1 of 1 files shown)


falkTX:
yes indeed, that somehow slipped through, thanks

PR #509, opened 2025-10-22
Wasm env add worker

dromer:
I needed this both locally and in CI, presumably because of `-j $(nproc)`


dromer:
ok nevermind. newer emsdk resolved it.

PR #512, opened 2025-11-02
Fix LADSPA hint calculation for logarithmic parameters

sboukortt:
For parameters from 1 to 65535 with defaults of 4096 and 128, the old logic picked “low” and “low”, resulting in default values of 16 for both, whereas this new logic picks “high” and “middle”, resulting in defaults of 4096 and 256.

### DIFF
--- a/distrho/src/DistrhoPluginLADSPA+DSSI.cpp
+++ b/distrho/src/DistrhoPluginLADSPA+DSSI.cpp
@@ -16,6 +16,8 @@
 
 #include "DistrhoPluginInternal.hpp"
 
+#include <cmath>
+
 #if DISTRHO_PLUGIN_WANT_PARAMETER_VALUE_CHANGE_REQUEST
 # error Cannot use parameter value change request with LADSPA or DSSI
 #endif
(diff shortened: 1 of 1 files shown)


distrho/src/DistrhoPluginLADSPA+DSSI.cpp
```diff
@@ -646,9 +646,21 @@ static const struct DescriptorInitializer
                     portRangeHints[port].HintDescriptor |= LADSPA_HINT_DEFAULT_MAXIMUM;
                 else
                 {
-                    const float middleValue =  ranges.min/2.0f + ranges.max/2.0f;
-                    const float middleLow   = (ranges.min/2.0f + middleValue/2.0f)/2.0f + middleValue/2.0f;
-                    const float middleHigh  = (ranges.max/2.0f + middleValue/2.0f)/2.0f + middleValue/2.0f;
+                    float middleValue;
+                    float middleLow;
+                    float middleHigh;
+                    if (hints | kParameterIsLogarithmic)
```
falkTX: syntax is not correct, this should be `hints & kParameterIsLogarithmic`

as you have it, it will always evaluate to true.

distrho/src/DistrhoPluginLADSPA+DSSI.cpp
```diff
@@ -646,9 +646,21 @@ static const struct DescriptorInitializer
...
+                    {
+                        middleValue = sqrt(ranges.min * ranges.max);
```
falkTX: The `sqrt` function uses double-precision, prefer to use `std::sqrt` here instead, so it automatically uses the most appropriate function variant.

distrho/src/DistrhoPluginLADSPA+DSSI.cpp
sboukortt: Oops, brain fart, that is indeed what I meant to write. Thanks.

distrho/src/DistrhoPluginLADSPA+DSSI.cpp
sboukortt: Thanks. Have I placed the `#include <cmath>` in the right spot or would you prefer somewhere else?

Issue #535, opened 2026-07-29
WebView on Linux when linking QtWebEngine: Some position related operation breaks after host window move.


HITOA:
Hi, I noticed an issue with dpf's webview about window position.

Some position related operation like drag & drop, and opening the context menu of the webview, start to break after moving the window.
This is immediatelly fixed after a resize, which seems to update whatever value was cached by the webview that wasn't updated by simply moving the window. I took a screenshoot to hopefully make the issue more understandable, The red dot is the position of the mouse, I can't screenshot my own mouse with my current setup;

<img width="476" height="356" alt="Image" src="https://github.com/user-attachments/assets/26ba8fdb-321b-4620-8e21-c9f2f8055ddc" />

Notice the offset between the mouse position when the right click was triggered and the position of the context menu. The offset between the two is exactly the offset between the current position and the last window position after resizing. The issue is the same with drag & drop. The rect for the drag zone is not updated correctly. which in my case is my main issue.

The fix seems to be about notifying the webview when the window position change. Calling ``webViewResize`` also seems to update whatever needed to be. I don't know if this bug happens on windows or mac, or even with Webkit GTK.

I'm on Linux, NixOS with wayland, The plugin should run with XWayland tho cause there doesn't seem to have any implementation of dpf webview for wayland anyway.  the plugin is linking at runtime to QtWebEngine ``libQt6WebEngineWidgets.so``.

falkTX:
This is already fixed in develop branch, did you try?

HITOA:
I just tried and pulled the last change on develop, yes sadly the issue still remain on my side even after a full clean build in case something was cached. Actually there is one more issue now the context menu doesn't even show up. It's not a big deal in my case but might be for some people.

HITOA:
Regarding the context menu issue, I'm not sure if I should post about it here, but just in case I thought this might be usefull:
```
Received signal 11 SEGV_MAPERR 000000000010
#0 0x7695f8d95da5 base::debug::CollectStackTrace()
#1 0x7695f8d81e98 base::debug::StackTrace::StackTrace()
#2 0x7695f8d9624a base::debug::(anonymous namespace)::StackDumpSignalHandler()
#3 0x769601c42790 (/nix/store/57iz36553175g3178pvxjij8z5rcsd4n-glibc-2.42-61/lib/libc.so.6+0x4278f)
#4 0x7695f8c3d114 

QWebEngineContextMenuRequest::isContentEditable()
[...]

PR #536, opened 2026-07-29
Fix webview position upon DragEnter event with QT WebEngine

HITOA:
Fix for #535 regarding webview positioning upon DragEnter event

### DIFF
--- a/distrho/extra/WebViewImpl.cpp
+++ b/distrho/extra/WebViewImpl.cpp
@@ -1069,6 +1069,7 @@ struct QEvent {
     uint8_t _[SIZEOF_QEvent];
     enum Type {
         Close = 19,
+        DragEnter = 60,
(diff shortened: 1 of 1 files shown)


distrho/extra/WebViewImpl.cpp
```diff
@@ -1850,6 +1851,17 @@ class WebViewEventFilter : public QWebChannelAbstractTransport
             return true;
         }
 
+        if (type == QEvent::DragEnter) {
+            if (_winId != 0)
```
falkTX: the winId check can be combined with the "if" above.
then add a comment explaining it, like:
> // HACK forcing webview window position, needed for drop area

falkTX (changes requested):

falkTX (approved):

